PR #22, opened 2017-04-11
Added hacky window scaling.

thomafred:
Quick window scaling implementation


Baekalfen:
Thanks for your interest in the project!

I've looked through the code, and I generally like the changes you made.

I like the way you reorganized the `Window_SDL2` file and the abstract class for the `GameWindow`, as we might see multiple implementations of the window.

I have a few points to go through:

1. When running the code as `pypy -OO main.py`, the window shows uninitialized memory and doesn't update the screen.
2. The performance hit is quite high. I think I will accept it as a temporary solution, if the performance hit is negligible, when running 1:1 scaling. My quick test currently shows 10-30% decreased performance for 1:1.*
3. I think the `SimpleFrameBuffer` should just be a part of `Window_SDL2` until it's used elsewhere.
4. Maybe rename `SimpleFrameBuffer` to something like `ScaledFrameBuffer`?
5. I would argue against the abstract class `FrameBuffer`, until we see a benefit of having multiple implementations.
6. You could rename `Window_SDL2` to `GameWindow_SDL2` or something like that, to reflect its base-class
7. Filenames have so-far been named in upper camel-case. I think you should do the same.

*I opened two terminals and navigated to each of our repos and ran `pypy main.py SDL2 ROMs/POKEMON\ BLUE.gb loadState` simultaneously. Then wait a minute for the JIT-compiler to warm up.

Just FYI, I'm currently working on hardware-accelerated scaling through SDL. Hopefully it will either improve performance, or at least add a small penalty independent on scaling factor. Just so we avoid doing the same work twice :)

thomafred:
Good Afternoon,

Thank you for you feedback.

I was initially working towards hardware-accelerated scaling myself, using OpenCL. This will was intended to be a post-processing step, utilizing some sort of [pixel-scaling algorith](https://en.wikipedia.org/wiki/Pixel_art_scaling_algorithms). 

This is where the *FrameBuffer*-classes come in. Any child class of *FrameBuffer* will act as a drop-in replacement for the *SimpleFrameBuffer*.

Adressing you remarks:
1. This is a bug I will look into.
2. I am aware of the perfomance hit, and is thinking of a moving the scaling/post-processing to a separate thread for the time being. Please see my notes on GPU post-rendering above.
3. I disagree with this. Even though *SimpleFrameBuffer* is only used by *Window_SDL2*, separating classes into files maintains readability.
4. This is a good idea, as this naming is more describtive.
5. See my arguments about post-processing above. Multiple *FrameBuffer*-implementations can also be nested.
6. This namechange would also be more descriptive considering the parent class.
7. I will keep this in mind. So far I have somewhat been sticking to [PEP8](https://www.python.org/dev/peps/pep-0008/]. If you have another coding-standard that you are adhering to, would you mind sharing this with me?

More commits to come :smile:

NB: I notice you are Danish. I am Norwegian myself, and have no trouble understanding Danish if this is easier for you.

/Tom

Baekalfen:
Looks very interesting with the pixel-scaling. That would be a really awesome feature to see in the emulator.

That's a good argument about `FrameBuffer`. But I'm unsure of the generality of it. I'll see how it looks, when I wrap up the hardware scaling. The `GameWindow` and `FrameBuffer` might be interferring with each other performance-wise. But we can include it for now, and then revisit.

2. We could make a new class of `FrameBuffer`, which is just a plain bitmap like it is now. Maybe call that one `SimpleFrameBuffer`? Then, when running in 1:1 scaling, we can simply choose the new class which doesn't apply any extra calculations for scaling like this: `self._screenBuffer = SimpleFrameBuffer(...) if scale == 1 else ScaledFrameBuffer(...)`
I can accept any performance hit for the scaling, as long as there is no alternate implementation. But maybe a small warning on stdout, if it's severe enough?
3. Let's include it for now.
5. Interesting thought about nesting. As long as it's add-on, we can do it. I just want to make sure, that the default settings are as fast as possible.
7. Generally I use PEP8 as inspiration, but I won't follow it at all cost. I'm fine with whatever is readable and sensible. The filenaming is just because it's replicating the class names inside. And then I just kept it for all files for some consistency.


I'll look forward to them. All constributions are highly appreciated 😊

PS.
Nice to meet a fellow Scandinavian 😄 I would like to train my Norwegian, but let's keep it international for other possible readers.

Baekalfen:
Hi Thomas,

I've just pushed two commits, which I think finalizes the pull request.

I made a slight change to the SimpleFrameBuffer to not use both _cache and _array. I know it worked as a latch for the buffer, but the performance hit was quite steep, and the synchronization I could see it providing, might be better put into the GameWindow implementation instead.

I also moved the FrameBuffer files into the GameWindow directory, to signal where it gets used.

Thanks again for your help, and please reopen the pull request, if there is any thing you are not satisfied with.

thomafred:
Thank you for approving my pull request. More to come.

To give you an idea of what I'm working on, check out my [feature branch](https://github.com/thomafred/PyBoy/tree/InputHandler)

PR #25, opened 2017-05-05
Reworked input handling

thomafred:
I have reworked the input handling completely, in preparation for loadable scripts.

Performance has worsened with this new input handler, hence the "WIP"-tag.


Baekalfen:
Wow, that's a big chunk of code! I'll try to look through it soon, but I might not have time today.

Could you give an overview of the changes and what the goal of the changes are?

Could you also make a simple test to roughly show the performance hit? I normally just load two instances side-by-side from different points in the commit-log and compare.

thomafred:
Firstly, I have moved the main event-loop out of ```main.py```. ```main.py``` no only initializes the window-object and use it to initialize a a ```PyBoy```-object. The event-loop is now in ```GbEvents/EventLoop.py``` in the form of a ```cycle```-method, which is called by the ```PyBoy```-object. This class uses ```GbEventLoop```, which is the main event-handler (some poor naming here).

The events have been given their own classes, which all inherit from the ```GbEvent```-class, located in ```GbEvents/GbEvent.py```. At the moment of writing, the events implemented are:

* __DebugToggleEvent__ : Toggle Debug-mode
* __FrameUpdateEvent__ : Used to signal a new frame is drawn and ready
* __InputEvent__ : User input such as key-press/release
* __MbTickEvent__ : MB tick
* __QuitEvent__ : Clean exit with optional save-state
* __SaveStateEvent__ : Load/Save MB state
* __SpeedChangedEvent__ : Emulation Speed changed

All event-implementations are located in ```GbEvent/Events/```.

The events are passed an abstract system state in the form of a ```GbSystemState```-object. This class holds some flags and "registers", used when interfacing with the emulated hardware.

All in all, the emulator should run as before, except with the performance-loss as previously mentioned. On my PC (Ubuntu 16.04 on VMware virtual machine), the emulator is running around 80-100%, peaking around 200% at load-state around "boot-time". Your master branch run around 80% steady, if memory serves me right.

I am pretty sure I know what causes this performance hit, so I will come back to the comparison after I have had a look at this.

thomafred:
After changing the ```GbButtonId```, the performance improved, and is on par with your master branch:



This test was carried out using *Pokemod Red* and ```DMB_ROM.bin``` with the following sequence:
1. Start game, emulator first loads GB firmware ```DMB_ROM.bin``` before loading the game
2. Let the game intro run until the title screen appears
3. Load a previously saved game
4. Walk a round around Prof. Oaks House
5. Exit emulator without saving

Baekalfen:
Performance looks fine, so that is not an issue.

I might not be seeing the bigger picture, but what do we gain from all these changes?

Maybe this should also be split into multiple pull requests? As I can see, it is not only the input handling that has been changed, but there is at least also a new logger?

Baekalfen:
You also say, that some of these changes are in "preparation for loadable scripts". What would these scripts be doing?

thomafred:
It is not only the input-handling that has been changed. All event-handling has changed. The most interesting part og this is of course the input-handling. As mentioned, this is in preparation for loadable scripts. The old event-handling did not really allow for much customization.

This new event-handling enables us to easily define new events and register them with the emulator. It also enables us to re-implement old events. 

The idea was that one can load ```PyBoy``` as a module , initialize a ```PyBoy```-object and the call the ```start()```-method. A separate thread may then trigger events with this object. For example, this thread can be running an A.I. that can be trained to play Mario, kind of like [SethBling's Mari.IO](https://www.youtube.com/watch?v=qv6UVOQ0F44).

In regards to the new logger: It is now a global instance, so there is no longer need to pass it as an argument to new objects, the way it was done before. The logger can also be configured to write to file.

The old logging-implementation did not support a I need I had when implementing the new event-handler, and was therefore included with this pull-request.

Baekalfen:
Could you demonstrate how you would implement an event? The ones already there, seems to deeply depend on functions implemented other places, so what would the capabilities of an event be?

I guess we could have PyBoy be loadable as a module, if anyone would need that. We could simply rename `main.py` into `PyBoy.py` and add the boilerplate stuff needed. But I sense, that you also think, that `main.py` has gotten too large, and I would agree on that.

About the AI. Wouldn't it be much easier to subclass `GameWindow_SDL2` or make a new `GameWindow` for this? This would give you access to all input events, the frame buffer, and even sprites, background and window separately.

I like the logger. I think it would be a good idea to split up this pull request, as I could would merge that in, without much difficulty.


On a note of code style. I don't see any reason to prefix all the new classes with `Gb`. Everything in the repo is about GameBoy, so it's a bit redundant to state.

thomafred:
The AI was just an example. The idea of loading scripts is that users may write short scripts to suite their purpose. Encasing this in an easy-to use sandbox-environment ensures that they do not need to go digging in the actual emulator code.

[I just pushed an event](https://github.com/thomafred/PyBoy/commit/097dd0aae52efda8b0cd6a229bb2b2fbaf236aa6) to do just this. This commit should demonstrate how to implement a new event with associated key-binding. The way event IDs are handled could be improved to make it even easier to make new events.

The loadable-scripts event is currently in its infancy and has some glaring flaws. Safety is one major issue that must be addressed. In the future, I will create a [sandbox-environment](https://wiki.python.org/moin/SandboxedPython) for loaded scripts to run in.

I will see if I can make the logger a separate pull-request.

On your note of the code style: I agree with you. The structure of my changes to the emulator needs some improvement. Expect a refactoring.

thomafred:
I do not think I will be doing any more work related to event-handling alone.

The "WIP"-tag has therefore been removed.

Baekalfen:
At the current time, I do not see all that much reason for such large changes, for this feature. I think we could achieve the same with a much more simple approach.

If you'd like, we could try to discuss the features you are missing, and how they could be implemented in a less invasive manner?

thomafred:
Sure.

In that regard, I would like to propose the following requirements for the reworked event-handling:

1. The event-handling schema should support a script-module at a later time.
2. The event-handling should be centralized and clearly defined.
3. The event-handling should be isolated from emulator internals such as ```MB```, ```RAM``` and ```AbstractGameWindow```-objects. Exceptions can be made for system-events if necessary.
4. Button-presses should be events.
5. Events can trigger other events. For example: A _button-event_ may trigger a _save-event_.
6. Event-handlers should be easily registered (added) and unregistered (removed) as needed. A sample use-case could be hot-plugging a controller, where the emulator may detect that the controller has been connected and register an input-event handler for this controller.
6. _Optional_: System-processes such as ```MB.tick()``` and display refresh should be separated into an event.

These are the requirements I had in mind when reworking the event-handling.

In addition to these requirement, we should clearly defined _master events_ that other events may inherit from. For example a master event ```InputEvent```:

Event Handling diagram

Baekalfen:
I think we agree on almost everything. But maybe we have different ideas of how much should be event-driven.

I see no benefit to run system events through an event-handler. It will be much easier to debug and follow around with function calls like it is now. And we might see some added overhead from the JIT not being able to see through it.

But I do see a great benefit to run keyboard- and joystick events through an event-handler like you describe. Maybe even add a signal for frame-updates and interrupts, but that would just be for the benefit of user-added eventhandler, not for the emulator.


For the code, I think it should be much more simplistic. I can see you added a range of classes and methods to make operations on them. Again, I might not see the bigger picture, in which case, you should make it clearer how we benefit from these.

I don't really have any coding style to point to (maybe I should make that?) but, I would recommend to put all files relevant to the event handling in one module, instead of `GbControls`, `GbEnum`, `GbEvent` and `GbInput` being in the main directory.

I really do appreciate the effort you are putting into this, so I hope you see this as a discussion, and not some strict rules to follow.

thomafred:
I apologize for my (very) late reply.

At the time of my previous reply, I went on holiday and immediately started a new job. After this I could not find the time to work on this.

I see you have implemented scripting support and other cool features since we last spoke. As a result, this PR has become redundant.

I might contribute in the future if I ever find the time.

Cheers.

Baekalfen:
No problem at all.

You're more than welcome to send in a pull request in the future :)

PR #37, opened 2018-03-30
Update py-sdl2 installation instructions

posixpascal:
Marcusva moved the py-sdl libraries to github lately:
https://github.com/marcusva/py-sdl2


Baekalfen:
Thanks!

Issue #33, opened 2017-11-20
Make a screen recorder


Baekalfen:
It would be useful to implement a feature in a GameWindow, which allows to record the game frame-by-frame. Contrary to regular screen-capture, this would allow perfectly timed frame-rates.

The output formats would preferably be GIF and MP4.

Issue #34, opened 2017-11-20
Run SDL2 in separate thread


Baekalfen:
For some games, nearly 50% of the CPU time goes into rendering the screen. This could easily be moved to another thread, to better utilize the available hardware.

Baekalfen:
Implemented in dfda9b6. Because of sluggish performance in PyPy's multiprocessing.Pipe, it is not advantageous yet. I expect it to improve in future releases of PyPy.

PR #38, opened 2018-06-24
Add screen recording functionality for GIFs

FKatenbrink:
This PR adds the functionality to save the screen as requested in #33. Currently, only the GIF format is supported. The default hotkey is _i_. It must be pressed to start and to end the capturing.

### DIFF
--- a/Source/PyBoy/GameWindow/FrameBuffer/framebuffer.py
+++ b/Source/PyBoy/GameWindow/FrameBuffer/framebuffer.py
@@ -36,6 +36,9 @@ class FrameBuffer(object):
         """Set FrameBuffer item at index"""
         self._cache[key] = item
 
+    def get_array(self):
+        return self._array.copy()
+
 class ScaledFrameBuffer(FrameBuffer):
 
     def update(self):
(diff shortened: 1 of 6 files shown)


FKatenbrink:
Thanks!

You're right, the saving procedure hangs the program as it takes quite a while to generate a 60 FPS GIF. To speed it up, we could try skipping every second frame and dropping the FPS to 30. Obviously, the quality would suffer.

I also tried running the saving procedure in a separate Thread to not block the emulation, however the saving procedure seems to take most of the CPU, so that the game is unplayable then. Any advice?

Baekalfen:
Personally, I can live with the hanging, as I will only use it occasionally.

I think it's best to leave the FPS at 60 as default, as this is what the Game Boy produces. In the future, we might see a settings window for the emulator, where the FPS could be changed, if the user wants to — alongside changing keyboard keys etc.

If you ran it as a thread, the GIL wouldn't be released, so emulator and saving procedure would be limited to one CPU core. To overcome this, you would have to use `multiprocessing.Process`, and transfer the video data to the separate process. Either it has to be transferred when saving, or frame-by-frame while recording. To transfer the memory, you can either use `multiprocessing.Pipe` to do a server-client style transfer or `multiprocessing.Array` to do it more implicitly.

Baekalfen:
I think you might be interested in the commit I just pushed. It should demonstrate all you need to move the saving to a separate process: dfda9b6. The commit was in reference to #34.

Baekalfen:
I updated your code, @FKatenbrink . Maybe you'll find it interesting :) d6e1a2f91897979e79818aa7702c184210d072cc

PR #44, opened 2018-11-15
Easier installation and more Git ignores

jardiacaj:
Hey, I wanted to try your emulator out and noticed some potential improvements on the first setup:

- IMHO having a requirements.txt file makes the setup easier.
- The imageio dependency was missing in the README.md (added it to requirements.txt).
- Added some more lines to the .gitignore file. These files were created by pip after installing the dependencies, IMHO it is convenient to let Git ignore them. I don't know if this is specific to my environment but I'd wonder if this hurt anyone.

Grats on the nice emulator! :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,3 +82,12 @@ ROMs/
 diffDump.txt
 Core_Dump_*
 *.state
+
+# Files from venv and dependencies
+Source/bin
+Source/include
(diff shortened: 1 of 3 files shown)


Baekalfen:
Thank you, and thanks for your contribution :)

PR #55, opened 2019-01-27
Python3-friendly imports and print()

stillinbeta:
PyPy3 doesn't actually work yet because marcusva/py-sdl2#132 is blocking upstream numpy, and numpypy doesn't work on python3. But once that issue is resolved, this should be most of what's needed for pypy3

### DIFF
--- a/Source/PyBoy/BotSupport/__init__.py
+++ b/Source/PyBoy/BotSupport/__init__.py
@@ -5,5 +5,5 @@
 # GitHub: https://github.com/Baekalfen/PyBoy
 #
 
-from Sprite import Sprite
-from TileView import TileView
+from .Sprite import Sprite
+from .TileView import TileView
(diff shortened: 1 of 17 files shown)


Baekalfen:
What is your recommendation? Should we merge it now, or should we wait for NumPy support?

Does the code work with 2.7 as before, if we added `from __future__ import print_function`?

stillinbeta:
The __future__ import isn't really necessary since we're only printing
strings. In Python 2.7, the parens are just interpreted as wrapping an
expression, so it still works there.

I'm testing this on 2.7 so all the changes are backwards compatible. I
think we may as well merge now, otherwise we may need to rebase a bunch
later.

Baekalfen:
Great, I've skimmed through the changes, tested it, and it works fine. I'll merge it now. Thanks for the help :)

PR #56, opened 2019-02-22
Add setup.py

stillinbeta:
I'm having a bit of trouble getting the dependency_link to work properly, but otherwise this works for me.

### DIFF
--- a/Source/setup.py
+++ b/Source/setup.py
@@ -7,7 +7,7 @@ setuptools.setup(
     name="PyBoy",
     version="0.0.1",
     author="Mads Ynddal",
-    author_email="mynddal@me.com",
+    author_email="mads-pyboy@ynddal.dk",
     long_description=long_description,
     content_type="text/markdown",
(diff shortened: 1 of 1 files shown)


Source/setup.py
```diff
@@ -0,0 +1,31 @@
...
+    author_email="mynddal@me.com",
+    long_description=long_description,
+    content_type="text/markdown",
+    url="https://github.com/Baekalfen/PyBoy",
+    packages=setuptools.find_packages(),
+    classifiers=[
+        "Programming Language :: Python :: Implementation :: PyPy",
+        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 2",
+        "License :: Freely Distributable",
```
stillinbeta: Creative Commons recommends [using a dedicated software license instead](https://creativecommons.org/faq/#can-i-apply-a-creative-commons-license-to-software).

Baekalfen:
That's awesome! It works fine for me too, so let's just merge it :)

PR #57, opened 2019-03-01
Window scaling in SDL2 and scanline-based GameWindow

### LINKED ISSUE #4: Hardware-accelerated window scaling in SDL
The current implementation if Window_SDL2.py does not work well with scaling, as it has a huge impact on framerate.

krs013:
Fixes #4 

Framebuffer objects in GameWindow are no longer needed, so I removed those.

On my way to understanding the code well enough to make the fix, I made my own GameWindow that renders a line during each call to `scanline` and scales and copies the buffer during VBLANK. It's still a little messy and could use a lot of optimization, but it's interesting and hopefully doesn't get in the way in the meantime.

I also made a few changes to the surrounding code to allow the scanline GameWindow to work, including passing the whole `lcd` object to the `scanline` method instead of just the viewports. I changed the other GameWindows to reflect that, and as far as I can tell.

Sorry for rolling all this extra stuff in, if a more atomic version would be preferred I can try to make a PR with only the changes to `GameWindow_SDL.py`, but hopefully this is alright!


Baekalfen:
Fantastic! I'll look through it, and merge as soon as I got some time. I hand in my thesis on Monday, so I'm a bit busy.

Baekalfen:
I had a look at it. It looks good, all-in-all. I had some minor changes, just to make the code slightly more intuitive. For example, explicitly importing sdl2, even though it might be imported implicitly. We never know, if some day it might change.

Then I moved three variables, from the object state, to where they are needed. This causes less confusion, as they have a more clear name, and their purpose is clearly shown, where they are used. Another developer (me), might not know what a variable is, if it's called `xs`, and is defined elsewhere. I renamed `dims`, just to match the other files with the same information.

The VSync function, was just to call it, if it is available. But we need a better way to limit the FPS.

I also moved two statements of `self._renderer.present()` to `updateDisplay`, as that is the functions intension. Again, just to match the other `GameWindow`s.

Baekalfen:
I saw you work on the regular SDL2 `GameWindow` as well. Really nice work! :)

I'll remove some comments, and merge your code.

krs013:
Those are good changes. I got a bit carried away when I renamed gameboyResolution to dims, trying to get those line lengths down to 79. At some point I may switch the `range` back to an `xrange` since it runs faster under Python2, but with any luck that won't matter for long. And thanks! I had fun working on this and look forward to the next task.

Baekalfen:
If it makes any difference, we can switch the `range` to `xrange`, but I doubt it makes a difference in PyPy. I think it's smarter than that. And as you say, with Py3, it won't make a difference.

Glad you like it! I appreciate your work.

Issue #16, opened 2016-11-13
Change readme instructions about pypy


Yardanico:
PyPy in ubuntu package repository is very old (5.1.2+dfsg-1~16.04 in xenial updates, and latest version is 5.6.0)
It's better to use Portable pypy binaries: https://github.com/squeaky-pl/portable-pypy  they can run almost on any linux distribution (including ubuntu) and does not depend on system libs)

Baekalfen:
Thanks for the heads-up. I'm not a regular Ubuntu user myself.

Would you like to correct it and make a pull-request? Otherwise I'll take a look at it.

TiberiumPY:
@Baekalfen english isn't my first language, but it will not that hard:
download pypy from https://github.com/squeaky-pl/portable-pypy  (and choose needed architecture)
extract it in any folder (better to have in home directory, like /home/tiberium/pypy)
Add /home/tiberium/oypy and /home/tiberium/pypy/bin to PATH
And then all your instructions.

TiberiumPY:
@Baekalfen by the way, there's already game boy emulator written in RPython (Restricted Python). PyPy is also written in RPython, so this emulator can be translated to C

Baekalfen:
Ok, but thanks anyway. I'll rewrite the Ubuntu/Linux instructions.

I have seen it before, but it seemed abandoned.. And we wanted to create our own emulator for the learning experience.

TiberiumPY:
@Baekalfen althrought it's abadoned, it fully functional

PR #59, opened 2019-03-07
Updated instructions for Pypy on Ubuntu/Linux

krs013:
Addresses #16 

I tested and updated the installation instructions for Pypy on Ubuntu. This was done on a pristine Ubuntu 18.04 LTS VM, so the dependencies are definitely enough to get things running.

This does not involve Portable Pypy, though I did test it and found that it could not install the dependencies because it wasn't configured by default to look at the system paths. It is still likely that alternate instructions could be provided for normal (not portable) installations of Pypy 7.0, or that we could figure out how to get Portable Pypy to compile against the system libraries (though that kindof defeats the purpose of it). However, I don't notice any speed difference between Pypy 5.1 (Linux) or Pypy 7.0 (macOS host). I also think it's beyond the scope of this project. Personally, I would rather try to eliminate or make optional imageio and numpy as dependencies so there wouldn't be a need for so many system libraries (although I'm not sure if that would fix things for libsdl2-dev).

Oh, and I changed the instructions to put the virtualenv in the root PyBoy folder instead of Source, because Source was feeling a bit cluttered and putting the venv in Source named it `(Source)` instead of `(PyBoy)` which is nicer.


krs013:
I'm not sure why it lists so many commits... could I have updated my fork wrong?

Baekalfen:
Hmm. I think you should have done some rebasing. But I think it is ok, if I squash and merge.

krs013:
That's what I get for leaving local changes on my master branch. I did a `reset --hard upstream/master` and will keep my work in other branches from now on so this shouldn't happen again.

Issue #60, opened 2019-03-07
Decouple GameWindow import from `main.py`


Baekalfen:
Currently, the `GameWindow` is imported from outside of the `PyBoy` module. It would de preferable, if it was selected with a flag through `sys.argv`, but the internals of PyBoy handled it.

krs013:
I believe this issue has been solved in the Cython update.

PR #62, opened 2019-03-11
Profile function calls, better argument parsing, and GameWindow decoupling

krs013:
This branch introduces profiling with cProfile. All function timings are tracked and reported, with some small overhead. In the course of implementing this, I also moved things around in `main.py`, using `argparse` to process command line arguments and removing references to `sys.argv` throughout the code. I also redid GameWindow selection and moved the functionality to `GameWindow/__init__.py`. The resulting code is cleaner in places, though some of the profiling code is a bit sprawling, and probably should be moved to its own class eventually. 

For now, though, maybe see if the functionality is good and if the argparse stuff looks okay. It will be better for running PyBoy as an API through other scripts to remove the direct `sys.argv` access.

The profiling code already helped me find an inefficiency in `GameWindow_Scanline.py`, and the fix for that is included here. Scanline is much faster now--almost as fast as SDL2.

### DIFF
--- a/Source/PyBoy/CPU/__init__.py
+++ b/Source/PyBoy/CPU/__init__.py
@@ -9,7 +9,7 @@ from ..OpcodeToName import CPU_COMMANDS, CPU_COMMANDS_EXT
 from .flags import flagZ, flagN, flagH, flagC
 from ..Logger import logger
 from .Interrupts import InterruptVector, NoInterrupt
-import numpy as np
+
 
 class CPU(object): # 'object' is important for property!!!
     from .opcodes import opcodes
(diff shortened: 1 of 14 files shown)


krs013:
Oh, and while this removes the previous profiling code that was used to measure the frequency of each opcode in the CPU, this framework can recreate that with the option `--profilingFilter opcodes.py`, and will also give the average time taken by each of those opcodes, if you do the math.

krs013:
Forgot to mention that this fixes #60

krs013:
To anyone wondering, this last commit and the first one are optimizations I found using the profiler, so it seemed fitting to add them to the PR. I’ll try to not clutter this too much more, though.

krs013:
These commits are after a rebase with the new changes--the earlier commits are gone but the result is the same (though now there is no conflict with generator.py and opcodes.py)

krs013:
This branch is no longer valid. I'm going to delete it and close the PR. Will probably open a new PR with the optimizations bc those are worth having, but the profiling stuff wasn't ready anyway.

PR #64, opened 2019-03-21
A couple optimizations

krs013:
This PR just contains a couple optimizations that make emulation a little faster (getSignedInt8) and make GameWindow_Scanline a *lot* faster (use of a buffer pointer instead of converting to a ctypes array every scanline). 

There's a bit of left over stuff in main.py where I changed how main was called, which could be dropped, but I'll probably revisit that area in my next PR anyway so I wouldn't worry about it unless it's bothersome.


Baekalfen:
Can we change the 255 and 128 to 0xFF and 0x80 respectively? I think it better conveys the message, that these are bitwise operations.

krs013:
To be honest I don’t think of them as bitwise operations, though I see what you mean. The key part is the addition and subtraction, as you overflow the number (mask to simulate 8-bit overflow) and then undo the offset, and addition and subtraction don’t really make sense as bitwise operations. 

Optimally, I’d write `((x+128) % 256) - 128`, but `%` is a bit slower so I changed it to `& 255`. `((x+128) & 0xFF) - 128` is the most correct, but it felt weird having two numeric types in an expression. I think using `0x80` for 128 would be less helpful, and it would be additionally unclear when combined with the +256 in the tile view selection to change -128 to +128. 

Unless I’m missing something, I think it’s best as is, but I could go for `((x+128) & 0xFF) - 128)` if that’s preferable to you.

Baekalfen:
Then I would prefer do the last one: `((x+128) & 0xFF) - 128)`. 0x80 should be apparent, as the middle value of an 8-bit number (7th bit set), but I get what you mean. But when we are masking out, then `0xFF` is definitely the way to go.

But on a different note, wouldn't it be better with: `(x ^ 0x80) - 128)`?

Baekalfen:
And can't we just delete to the whole `MathUint8.py` file?

krs013:
Oh, yeah, I totally missed that, xor and subtract is much better. Glad you noticed that! And yeah I forgot to check if MathUint8.py was used elsewhere too. Thanks for the fixes.

PR #66, opened 2019-03-31
Moved include management to makefile and cleaned it up a bit

krs013:
Works on macOS and Ubuntu 18.04.


krs013:
Oops, I forgot that this branch wasn't sticking around--thanks for copying it over.

PR #70, opened 2019-04-06
No NumPy (and a Window_SDL2 that supports PyPy again)

### LINKED ISSUE #67: Better detection of compiled/interpreted/JIT backend
The SDL2 `Window` doesn't work in CPython without Cython, as we cannot differentiate it from the Cython version.

krs013:
Fixes #67 

The ScreenRecorder doesn't work with this commit, but I'm working on a replacement that is mostly functional.

This PR also includes a bug fix for Window_SDL2, where sprites close to the top of the screen did not display correctly, and removes a lot of commented code that IMO belongs in the git history for now.


krs013:
Updated with four commits that repair screen recording for the SDL2 window. In keeping with the no-numpy theme, the updated ScreenRecorder uses `Pillow` instead of `imageio` (which depends on `Pillow` anyway). It is feature-equivalent to the old version, runs very fast under Cython, and is still fast enough to work under PyPy without slowing down emulation past 100%.

PR #79, opened 2019-09-04
Fixed identifiers and added some error handling in save/load state

Baekalfen:


Source/pyboy/botsupport/tile.py
```diff
@@ -49,18 +49,23 @@ def __init__(self, mb, identifier=None, index=None):
         if index is not None:
             assert isinstance(index, tuple), "Index parameters needs to be a tuple of (int:index, bool:signed)"
             index, signed = index
-            assert 0 <= index <= 0xFF, "index out of range"
             assert isinstance(signed, bool), "signed is not instance of bool"
+            if signed:
+                assert -128 <= index <= 127, "index out of range"
+            else:
+                assert 0 <= index < LOW_TILEDATA_LENGTH, "index out of range"
```
krs013: This could be cleaner. `<128` instead of `<=127` to be consistent, and `LOW_TILEDATA_LENGTH` should be in bytes like the rest of the constants unless its name is explicit, like `LOW_TILEDATA_NTILES` or something (and use literal numbers instead of a constant like the surrounding code).

Baekalfen: I had a lot of trouble finding a good name for the var. But `LOW_TILEDATA_NTILES` could be a good name. It's used several places, that's why I didn't use a literal. 

I get you with the `<=` and `<`. I thought the range of a signed int8 is usually written that way. And the the `<` is because the constant is used other places. I didn't want to do `<= const-1`

krs013: Most of the places where `LOW_TILEDATA_LENGTH` was used as a lone number, it was among literal numbers, and I think seeing 256 with 128 and 384 isn’t out of place. Where it was used as a memory size, it was used along with other constants, but we had to multiply by 16 (size of a tile, but that’s maybe less obvious than the size of a byte). So my preference would actually be to just use 256 where we need the number, and make `LOW_TILEDATA_LENGTH` (or maybe `SIZE`) the size of the table in bytes, 16*256.

krs013: Most of the places where `LOW_TILEDATA_LENGTH` was used as a lone number, it was among literal numbers, and I think seeing 256 with 128 and 384 isn’t out of place. Where it was used as a memory size, it was used along with other constants, but we had to multiply by 16 (size of a tile, but that’s maybe less obvious than the size of a byte). So my preference would actually be to just use 256 where we need the number, and make `LOW_TILEDATA_LENGTH` (or maybe `SIZE`) the size of the table in bytes, 16*256.

Baekalfen: But if you look at this line, it will need some explanation, if we just add 256 out of the blue: https://github.com/Baekalfen/PyBoy/pull/79/files#diff-64ece5aedc106e4d69976e53cacd6ad1R130

krs013: Sure, but the -128+256 evaluates to +128 and there you could just leave a comment explaining the whole thing

Baekalfen: I see your point, and I do agree with you, that it would be a good approach as well. But it's not directly misleading ATM, so I'll just leave it be.

PR #80, opened 2019-09-07
Fixed high-to-low interrupt to trigger correctly

Baekalfen:


Source/pyboy/interaction.py
```diff
@@ -58,6 +60,11 @@ def key_event(self, key):
         elif key == windowevent.RELEASE_BUTTON_START:
             self.standard = set_bit(self.standard, P13)
 
+        # XOR to find the changed bits, AND it to see if it was high before.
+        # Test for both directional and standard buttons.
+        return ((_directional ^ self.directional) & _directional) and \
```
krs013:
```suggestion
        return ((_directional ^ self.directional) & _directional) or \
```

Baekalfen: Yes, my mistake

PR #86, opened 2019-10-26
Refined testing and automatic GIF generation

Baekalfen:


.flake8
```diff
@@ -14,4 +14,7 @@
 ignore = E261, E123, E221, E203, W503, W504, E226, E722, E126, E121, E731, E402
 per-file-ignores = interface_example.py:E501, tetris_bot.py:E501
 exclude = .git, __pycache__, opcodes.py, __init__.py
+per-file-ignores =
```
krs013: This is a duplicate of line 15

Baekalfen: Damn, you're right. Then I think we should remove the first line, as it didn't work.

krs013:
Oh, by the way. After merging, `README.md` probably should be edited as well, since the different gifs have slightly different `width`s.

Issue #87, opened 2019-10-31
Road to PyPI


krs013:
This issue is for tracking the tasks that need to or should be done before putting the project up on PyPI.

- [x] Restructure source tree: put `pyboy` directory at the root level, and move scripts like `main.py` inside, possibly as `__main__.py` (`python -m pyboy`) or as a `setuptools` script.
- [x] Move vital functionality into `setup.py` so it doesn't need external configuration (currently in `Makefile`) to run properly.
- [x] Figure out how to deal with the Cython/PyPy duality.
- [x] Figure out what goes in the `sdist`: Cython? Cythonized `.c` files? Test modules? Examples? OpenGL w/o dependencies?
- [x] Figure out how to make an `sdist` and get `setup.py` to install from it via pip.
- [x] Ensure correctness of `requirements.txt`. (#82)
- [x] ~~Speaking of bundling, figure out how to bundle SDL2 or maybe download a copy on install? Maybe that's for built distributions, though.~~ Nah
- [x] Fix attribution for `sdl2.pxd`, or install as a dependency.
- [x] Ensure that PyBoy can run outside of the build directory structure (i.e., won't crash for lack of directories like ROMs, recordings, etc...
- [x] Silence the logging printouts, or at least add an option to do so. (#88)
- [x] Assign a semantic version number.
- [x] Double check license compatibility.
- [x] Claim `pyboy` on pypi.org and upload PyBoy!

That's all I can think of for now but there are probably things I forgot. Also we can mention issues here that we'd like to fix before uploading, but I don't think any of the current ones besides #82 need to be closed before we have something publishable. We can discuss the items in the checklist here or create separate issues for the conversation.

Baekalfen:
Nice job. Looks like you did your homework ;)

About bundling system dependencies (SDL2/OpenGL), from experience, that is just not handled by `pip`. For example with NumPy, PyOpenGL, and PySDL2, they all just crash, if the system dependencies are not installed.

krs013:
Yeah, looks like it's not canonical, but I think it'd be technically possible to bundle a copy to fall back on if there's no system version. Even if we do that eventually, though, we can safely choose to not bother for now, so I'll update the list to reflect that.

a-hurst:
[...]

PR #90, opened 2019-12-18
Support native Windows Python compilation with setup.py rather than make

krs013:
With these changes, PyBoy can be compiled with Microsoft's Visual Studio build tools, instead of requiring mingw and using make, it is invoked with `python setup.py build_ext -i` directly.

Full details in the commit message for 6d647ea.

The first commit is rather standalone so I left it alone, but the second is a squash of several commits that can be viewed individually on krs013/PyBoy:windows-native-workingcopy, but there's nothing really interesting there.

I have not run `make test` on this branch successfully bc my setup differs too much to run it without changing the `Makefile`. (I use venvs so CPython and PyPy can't run from the same session. I have a way to fix it but that can come in a different PR). That said, there's no reason for the tests to fail because of these changes, as very little of PyBoy itself is involved.

With the update to `.pre-commit-config.yaml`, you might have to update `pre-commit` with `pip install -U pre-commit`, but maybe not. I never checked if my version matched the config before.

EDIT: Oh, I should mention this fulfills item 2 in #87.


krs013:
Huh, forgot to update the README to indicate that Windows is supported too. We'll have to get that later.

PR #94, opened 2019-12-22
Support both PyPy and Cython in setup.py

Baekalfen:


setup.py
```diff
@@ -1,26 +1,63 @@
...
 import sys
 from distutils.command.clean import clean as _clean
 from distutils.command.clean import log
 from distutils.dir_util import remove_tree
 from multiprocessing import cpu_count
 
 from setuptools import Extension, find_packages, setup
+from setuptools.command.test import test
+
+CYTHON = platform.python_implementation() == "PyPy"
```
krs013: Why is `CYTHON` True if the implementation is `PyPy`? Isn't that backwards?

setup.py
```diff
@@ -109,77 +150,77 @@ def define_lib_includes_cflags():
     return libs, libdirs, includes, cflags
 
 
-# Cython doesn't trigger a recompile on .py files, where only the .pxd file has changed. So we fix this here.
-def touch_changed_py_files():
+def prep_pxd_py_files():
```
krs013: This is awesome!

setup.py
Baekalfen: Yep, it's backwards. It's fixed locally on my uncommited changes.

Issue #89, opened 2019-12-13
Replacement Boot ROM fails to start Pokemon Gold


krs013:
When using the fallback boot ROM with Pokemon Gold, the game doesn't seem to load at all, as the screen stays completely blank and the usual MBC warnings don't print out after the usual amount of time.

In case this can't be easily fixed, possible mitigations include adding an extra warning to the existing message that is printed when using the fallback boot ROM and adding the fallback ROM to the test suite.

Otherwise, we might be able to find out what is trapping execution for the game and fix this behavior. It may just be that GBC games require an extra step of authentication or palette/clock initialization before they can run.

krs013:
NB: It was also pointed out on Discord that the fallback BIOS causes a segfault with Pokemon Gold on windows under minGW, but I haven't been able to reproduce this yet.

Baekalfen:
The way to solve this, is to find the exact state the original boot ROM leaves the Game Boy in, and make the replacement ROM to reconstruct the same state. There are a lot of things, that the replacement currently doesn't touch.

Baekalfen:
It would be awesome, if we could get the replacement to show a PyBoy logo on boot ;)

PR #98, opened 2019-12-30
PyBoy boot ROM

Baekalfen:
Solves #89.

The logo it shows on boot is pretty rough, but can easily be changed with the provided scripts and Makefile.

It also initialize the window classes with a white background, as the GameBoy would. Otherwise, the boot logo will be on a black background.


krs013:
`bootrom.bin` does not seem to be a part of this PR, but I think you meant to put it in.

Also, even though the default boot-ROM is now our replacement, the warning message still prints out if the `DMG_ROM.bin` rom isn't found (even though it's not being used). That message should probably just be removed at this point.

.flake8
```diff
@@ -1,21 +1,27 @@
...
-# E128: flake8 is either buggy, or the rule is nonsensical: Continuation line under-indented for visual indent
-# E221: Sometimes alignment of operators is better. Use common sense: "multiple spaces before operator"
-# E203: Sometimes alignment of commas is better. Use common sense: "whitespace before ','"
-# W503: Sometimes common sense prevail: "line break before binary operator"
-# W504: Sometimes common sense prevail: "line break after binary operator"
-# E226: "If operators with different priorities are used, __consider__ adding whitespace around the operators with the lowest priority(ies). Use your own judgment;..."
-# E722: Use common sense: "do not use bare 'except'"
-# E126: Do what feels right
-# E121: Do what feels right
-# E731: "E731 do not assign a lambda expression, use a def" -- don't tell me how to use lambdas
```
Baekalfen: I kinda liked the snarky comments ;)

bootrom/Makefile
```diff
@@ -3,20 +3,12 @@
 
 all: build
 
-PY := python3
```
Baekalfen: I had these to reflect consistency between the Makefiles. From what I've seen, this is also the norm.

bootrom/Makefile
```diff
@@ -3,20 +3,12 @@
...
-ROOT_DIR := $(shell git rev-parse --show-toplevel)
```
Baekalfen: The purpose was to exactly tell where the output was going, even if the `bootrom` directory moved.

bootrom/logo_to_tiles.py
```diff
@@ -0,0 +1,72 @@
...
+asm_out = ".logo:\n"
+for name in ["P1", "P2", "Y1", "B2", "O", "Y2"]:  # Order is important, so we can't iterate on TILES
+    rows = [int(f"0b{row.replace('.', '0').replace('x', '1')}", base=2) for row in TILES[name].split()]
+    asm_out += f".{name}\n"
+    asm_out += "    DB "
+    line = []
+    for row in rows:
+        line.append(f"${row:02x}")
+    asm_out += ', '.join(line)
+    asm_out += '\n'
```
Baekalfen: Nice rewrite 👍

bootrom/Makefile
krs013: That’s fair. I had some ideas about compatibility when I changed it but really there was no reason for me to remove them so it’s good to put it back.

.flake8
krs013: Haha it was funny at first but I kept losing track of what warning was which... also I’ve wanted to sort them for a while anyway.

bootrom/Makefile
krs013: Technically the only python is python3 now ;)

krs013:
So I have a version of this that uses italics but is a little less slanted, but all of these only display the logo for one frame. Not sure why. The scrolling one is cool too and I might like that better, but the logo disappears once it’s done moving on that one too. Once we can fix that I think this is ready to merge.

bootrom/logo_to_tiles.py
krs013: Thanks! It was pretty fun. String literals are excellent for this kind of thing.

PR #102, opened 2020-01-14
Fix setup.py install_requires calculation

krs013:
There was a small error in the calculation of `install_requires` in `setup.py` that caused cython to be included only if using PyPy, which is why `pip install` didn't seem to be sufficient for installing dependencies. This PR fixes that and also cleans up an extra line in `setup.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -252,7 +252,7 @@ setup(
         ],
     },
     cmdclass={'build_ext': build_ext, 'clean': clean, 'test': PyTest},
-    install_requires=([] if CYTHON else ["cython"]) + [
+    install_requires=(["cython"] if CYTHON else []) + [
         "pysdl2",
         "numpy",
         "Pillow",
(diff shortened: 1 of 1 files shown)


krs013:
Actually, there is a deeper problem that this commit does not fix (although the commit is still good). The preamble to `setup.py` imports a number of packages from Cython, but it seems that `pip` tries to run `setup.py` before installing dependencies (given that the dependencies are actually specified by `setup.py`). So we have a bit of a circular dependency to resolve. I'm not sure exactly how to do this.

Baekalfen:
Are you sure this is correct? The logic is:
```
CYTHON = platform.python_implementation() != "PyPy"
["cython"] if CYTHON else []
```

Which means, CYTHON is true, if we are not in PyPy, and then add "cython" to the list, if CYTHON is true. So: Add cython, if not PyPy.

krs013:
Yes, that's what I'm changing it to. The previous/current version is `([] if CYTHON else ["cython"])`, which is a holdover from when `CYTHON` was negated.

Baekalfen:
Oh, you're totally right!

PR #103, opened 2020-01-18
add requierments file

mralext20:


Baekalfen:
There's a spelling mistake in the filename :)

I think we should remove the version pinning, as long as we don't have a specific reason to use an exact version.

Could this be unified with the `setup.py` file? Maybe the `setup.py` file could read from the `requirements.txt` and populate `install_requires` and `extras_require`?

PR #101, opened 2020-01-13
Add Dockerfile

mralext20:

PR #104, opened 2020-01-18
Reopening mralext20's Docker PR

Baekalfen:
See #101 for more details


krs013:
The Dockerfiles seem to be pretty stable now, and functionally the PR is viable, but it would be good for the tests to be working on the Docker images for coverage's sake.

Since certain ROMs/packages/directories are required to run the tests, it may be useful to add a "pretest" command to `setup.py` (or maybe just a step to the tests themselves) that checks for the presence of the game ROMs, test ROMs, and output directories needed to run the tests themselves. (Though I imagine that the tests writing the Blargg output to `../PyBoy.wiki/Blargg.md` is just leftover from when the wiki used to feature that report). This doesn't naturally relate to the Dockerfiles, but it seems like it might be helpful in debugging the tests as we develop a process to test the images without actually putting ROMs in the images.

Another thing we might do while we wait is consider whether to make an extra Dockerfile or two for running PyBoy under PyPy instead of CPython. If we do that, we should move all the Dockerfiles to a `dockerfiles/` directory so we don't clutter the root directory too much. It would be nice to do though, especially if we hope to automate testing in the future with Docker.

krs013:
Also we need to add a `.dockerignore` file. At the very least, it should prevent any of us from accidentally bundling ROMs into published images, which could cause problems. It would also save us from wasting image space on `.git` history and things like the readme gifs and report pdf.

Baekalfen:
Good idea about the `.dockerignore`! The other things, I'm already working on, on my local branch. I wanted to make the tests pass before merging. Otherwise, it's too hard to maintain.

Dockerfile.ubuntu1804
```diff
@@ -0,0 +1,20 @@
+FROM ubuntu:18.04
+
+RUN apt-get update && apt-get install -y \
+  build-essential \
+  libsdl2-dev \
+  python3 \
+  python3-pip \
+  python3-dev \
+  git \
+  valgrind vim \
```
krs013: I'm assuming these are left over from some debugging? Probably not needed.

PR #105, opened 2020-01-20
use the docker cache, it makes things faster

mralext20:


mralext20:
this requires a requirements.txt.

PR #110, opened 2020-02-15
Rewrite MBCs and add MBC5

nicoeps:
Rewritten MBC implementation to follow Gekkio's documentation: "Game Boy: Complete Technical Reference".
Passes all Mooneye MBC1 and MBC5 tests, but not the multi cart test.
Added MBC5.
Fixed minor things in base_mbc.py and cartridge.py.


pyboy/core/cartridge/mbc1.py
```diff
@@ -11,28 +11,14 @@
...
-            if (value & 0b00001111) == 0x0A:
-                self.rambank_enabled = True
-            else:
-                self.rambank_enabled = False
+            self.rambank_enabled = (value & 0b00001111) == 0b1010
         elif 0x2000 <= address < 0x4000:
-            # But (when using the register below to specify the upper ROM Bank bits), the same
-            # happens for Bank 20h, 40h, and 60h. Any attempt to address these ROM Banks will
-            # select Bank 21h, 41h, and 61h instead.
+            value &= 0x1F
```
Baekalfen: Maybe write `0x1F` as bits to make it perfectly clear: `0b00011111`? Generally I use hex if it's one of: 0x00, 0xF0, 0x0F, 0xFF, 0x80 and mostly any other mask as bits.

pyboy/core/cartridge/mbc1.py
```diff
@@ -11,28 +11,14 @@
...
         if 0x0000 <= address < 0x2000:
-            if (value & 0b00001111) == 0x0A:
-                self.rambank_enabled = True
-            else:
-                self.rambank_enabled = False
+            self.rambank_enabled = (value & 0b00001111) == 0b1010
         elif 0x2000 <= address < 0x4000:
-            # But (when using the register below to specify the upper ROM Bank bits), the same
-            # happens for Bank 20h, 40h, and 60h. Any attempt to address these ROM Banks will
-            # select Bank 21h, 41h, and 61h instead.
```
Baekalfen: I think a version of the comment should stay. It is not obvious why 0 should become 1 from reading the code alone

pyboy/core/cartridge/mbc1.py
```diff
@@ -11,28 +11,14 @@
...
+            value &= 0x1F
             if value == 0:
                 value = 1
-            # sets 5LSB of ROM bank address
-            self.rombank_selected = self.rombank_selected & 0b11100000 | value & 0b00011111
+            self.bank1 = value
```
Baekalfen: Wouldn't it have worked with the old code? It seems like the variables `self.bank1` and `self.bank2` could be avoided?

pyboy/core/cartridge/mbc1.py
```diff
@@ -11,28 +11,14 @@
...
         elif 0x4000 <= address < 0x6000:
-            # Note: 16Mbit = 2MB. 2MB/(8KB banks) = 128 banks. 128 is addressable with 7 bits
-            if self.memorymodel == 0: # 16/8 mode
-                # sets 2MSB of ROM bank address
-                self.rombank_selected = self.rombank_selected & 0b00011111 | (address & 0b11) << 5
-            # Note: 4Mbit = 0.5MB. 0.5MB/(8KB banks) = 32 banks. 32 is addressably with 5 bits
-            elif self.memorymodel == 1: # 4/32 mode
-                self.rambank_selected = value & 0b00000011
```
Baekalfen: It this functionality supported somewhere else, or was it a wrong implementation?

pyboy/core/cartridge/mbc1.py
```diff
@@ -11,28 +11,14 @@
...
-            else:
-                raise logger.error("Invalid memory model: %s" % self.memorymodel)
+           self.bank2 = value & 0x3
```
Baekalfen: I think bitmasks are more clear (`0b11`), when we are explicitly filtering for specific bits.

pyboy/core/cartridge/mbc1.py
```diff
@@ -42,6 +28,28 @@ def setitem(self, address, value):
                                "banks are not initialized. Initializing %d RAM banks as "
                                "precaution" % (value, address, EXTERNAL_RAM_TABLE[0x02]))
                 self.init_rambanks(EXTERNAL_RAM_TABLE[0x02])
-            self.rambanks[self.rambank_selected][address-0xA000] = value
+
+            if self.rambank_enabled:
+                self.rambank_selected = self.bank2 if self.memorymodel else 0
```
Baekalfen: Wait, so the RAM bank is changed by the same value as the ROM bank?

pyboy/core/cartridge/mbc1.py
```diff
@@ -42,6 +28,28 @@ def setitem(self, address, value):
...
+                self.rambanks[self.rambank_selected % len(self.rambanks)][address-0xA000] = value
         else:
             raise logger.error("Invalid writing address: %s" % hex(address))
+
+    def getitem(self, address):
+        if 0x0000 <= address < 0x4000:
+            self.rombank_selected = (self.bank2 << 5) if self.memorymodel else 0
```
Baekalfen: I like being explicit about boolean conversions. It tells more about your intentions. What value do you assume `self.memorymodel` is? Should it be `self.memorymodel != 0`, `self.memorymodel > 0` or always `self.memorymodel == 1`?

pyboy/core/cartridge/base_mbc.py
```diff
@@ -30,10 +30,11 @@ def __init__(self, filename, rombanks, external_ram_count, carttype, sram, batte
         self.gamename = self.getgamename(rombanks)
 
         self.memorymodel = 0
+        self.bank1 = 1
+        self.bank2 = 0
```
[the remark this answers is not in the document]
Baekalfen: `bank1` and `bank2` are only used in MBC1, so I think you should move it to that class

PR #111, opened 2020-02-16
Correcting cross-call calculate_cycles cycle counting

krs013:
This PR contains two commits so far. The first is an adjustment to the method `calculate_cycles` that keeps track of how many cycles it runs across calls to account for overshoot when an instruction takes more cycles than the target number. We actually devised these changes almost a year ago but it got lost in a stale branch... oops.

The second commit consists of changes to the test cases that were required to reset the replays and get the regression tests passing--this is warranted because the test failures are due to desired changes to emulation.

All tests pass on macOS and Linux under Cython. I can't get all tests to pass on PyPy, though I don't think it's related to the changes (it may be related to the test library changes). Previous commits fail some tests for me as well under PyPy, but that could be my setup.

If we can determine what is wrong with the PyPy tests, it might be worth fixing before merge. I would still like to try making some changes to the test library in the future, probably using `tox`, and PyPy would receive more attention then. Another subject of focus would be simplifying the process of updating tests when a desired change is made. I considered trying to do that in this PR but then we'd never merge in the actual correction! But the sooner we fix this, the better, and this is necessary for sound anyway.

It should be noted that changes like this will probably affect downstream projects that rely on frame-level consistency like bots, TAS bots, or some AI's. We'll probably want to include that warning in the merge commit message.


Baekalfen:
Nice with the reset flag in the tests!

krs013:
Ah, nice, the cycles_remaining is definitely neater. It may affect sound calculation, since that needs to know when register writes happen at the clock level, but there may be a better solution for that anyway. We'll see.

PR #112, opened 2020-02-21
Add get_scanline_parameters method

MisterKeefe:
Hi, 

Not sure if this is generally useful -- this PR adds a method to retrieve the `_scanlineparameters` from the window object. It's implemented only for the `SDLWindow` and children, and returns a list of lists, one per scanline, with the position(s) in the tilemap used for each scanline per frame. 

From experimentation, this seems to work well. There's a comment on `get_screen_position` with a TODO about retrieving the scanline parameters instead -- not sure if this meets that requirement. 

Happy to flesh this out a bit to get it up to scratch!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -135,3 +135,4 @@ bootrom/pyboy.png
 bootrom/PYBOY_ROM.bin
 
 pyboy.tar.gz
+test.replay
--- a/pyboy/pyboy.py
+++ b/pyboy/pyboy.py
@@ -444,6 +444,10 @@ class PyBoy:
         """
         return (self.mb.lcd.getviewport(), self.mb.lcd.getwindowpos())
 
+    def get_scanline_parameters(self):
+        """Return the scanline parameters"""
+        return self.window.get_scanline_parameters()
+
     def save_state(self, file_like_object):
         """
         Saves the complete state of the emulator. It can be called at any time, and enable you to revert any progress in
--- a/pyboy/pyboy.py
+++ b/pyboy/pyboy.py
@@ -428,7 +428,6 @@ class PyBoy:
         """
         return botsupport.TileMap(self.mb, window=True)
 
-    # TODO: We actually need the scanline_parameters. Mario might always report (0,0) at the end of frame.
     def get_screen_position(self):
         """
         These coordinates define the offset in the tile map from where the top-left corner of the screen is place. Note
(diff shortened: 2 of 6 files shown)


Baekalfen:
Really great work! I can see you've taken care of everything! @krs013 how does your plan look for the `scanline_parameters` will they survive your refactoring? Or should we include it in the API even though it's not used in the renderer?

krs013:
Hard to say at this point. If it's something that would be needed for user programs on some games then we can make sure the information is available--even if not in this form. Since we'll want to keep the debug view working at least as well as it is now, it's likely that the scanline parameters will be recorded one way or another. 

Would it be more useful to let the user inspect the screen data and be able to track it back to its source in the tile maps? Probably. Even in that case, I could see a reason to want to be able to track the scanline parameters on a per-line basis. So it's probably safe to merge this, and expect the name or data format to change slightly in the future, but anything written with may have to change slightly.

Also, is there a reason you chose to go with `np.vstack` instead of just passing the whole array to `np.array`? `np.array` can process nested lists properly, unless you found it didn't respond well in Cython.

Baekalfen:
You're right. Even just for the debugging view, we will always need some functionality to record this. I'll hopefully get time to add some tests later today (European time), and get it merged.

MisterKeefe:
The reason for the `vstack` is that when I was just passing the entire thing in (IE `np.array(self._scanlineparameters)`) numpy seemed to have trouble with recognising that it was a 2D array of 4 bytes per row, instead reading it as a flat array. This seemed to be true regardless of what `dtype` I passed in. 

That said, there has to be a way of just reading `uint8_t[144][4]` straight into a numpy array.

Baekalfen:
How about `.reshape(144, 4)`?

MisterKeefe:
The problem seems to be that `self._scanlineparameters` is converted by cython into a list of bytestrings when it's passed into `numpy` functions. Since `list` doesn't expose a buffer interface, numpy can't read this object all in one go. 

Something like: 

```
np.frombuffer(np.array(self._scanlineparameters, ...), ...) 
```

also doesn't work because the null bytes in `self._scanlineparameters` cause them to be evaluated (in python land) as empty bytestrings, which `numpy` then complains about converting into array entries.

The approach currently in the PR does work (although it's probably quite slow compared to what numpy/cython are truly capable of) because it explicitly pulls out individual values from the array and promotes them to python `ints` before handing them off to numpy. 

I think there are fixes for this, but they might involve a lot more extensive modifications to the `SDLWindow` than currently.

Baekalfen:
I tried optimizing it a bit, but Cython is really not doing its job of cooperating.. I've left your solution, as it is the most sensible one, and added some tests. The feature is more important than the performance, so we'll revisit it some other time.

Baekalfen:
And thanks again for you work!

PR #117, opened 2020-04-19
Update README.md

vyscond:
- Covering Fedora steps to install SDL2 lib
- Recommending pip installation with `--user` flag

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,12 +23,11 @@ Installation
 The instructions are simple, if you already have a functioning Python environment on your machine.
 
  1. Install SDL2 through your package manager:
-
  - Ubuntu `sudo apt install libsdl2-dev`
  - Fedora `sudo dnf install SDL2-devel`
  - OS X `brew install sdl2`
 
- 2. Install PyBoy using __`pip install --user pyboy`__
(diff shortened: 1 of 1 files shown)


Baekalfen:
The `--user` might be subjective. I get the following warning:
```
WARNING: The script pyboy is installed in '/Users/mads/Library/Python/3.7/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  NOTE: The current PATH contains path(s) starting with `~`, which may not be expanded by all applications.
```

It that a general recommendation?

krs013:
I would consider `--user` to be good practice, though perhaps not as good as using a virtual environment. If it's the only alternative to `sudo pip3 install pyboy` on a system running Ubuntu or Debian (can't test atm) then `--user` is infinitely better. The only thing lost is the ability to run PyBoy as a script "`pyboy`" instead of the longer `python3 -m pyboy`, and we can add a note in the wiki about adding `~/Library/Python/3.7/bin` to `PATH` if that behavior is desired.

Baekalfen:
Thanks for the help @vyscond ! I've chosen some middle ground, as out of my test systems, only  Windows requires it (Mac, Ubuntu, Manjaro and Windows)

PR #120, opened 2020-04-21
Implemented memory override method for PyBoy object (currently only M…

grungi-ankhfire:
…BC1)

This is for your consideration, it is definitely not done (it should at the very least work with more MBC types), but is this more or less what you had in mind?

I also made it so for now you can only override the loaded rom banks, but it would possibly be more interesting to let people change any value anywhere in the rom? What is your opinion on that?

### DIFF
--- a/pyboy/core/cartridge/base_mbc.pxd
+++ b/pyboy/core/cartridge/base_mbc.pxd
@@ -34,7 +34,7 @@ cdef class BaseMBC:
 
     cdef uint8_t getitem(self, uint16_t)
     cdef void setitem(self, uint16_t, uint8_t)
-    cdef void overrideitem(self, uint16_t, uint8_t)
+    cdef void overrideitem(self, int, uint16_t, uint8_t)
 
 
 cdef class ROMOnly(BaseMBC):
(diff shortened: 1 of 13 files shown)


Baekalfen:
Clever to do it at the ROM level. I think I need to think this through before we go and add it to the API. Once people start using something, it's hard to go back and change it.

Would it be useful to make a RAM value read-only with the same function? Like forcing the memory address for cash to stay at "9999".

Baekalfen:
For ROM overriding, we could take a ROM-bank parameter, to specify any bank.

Baekalfen:
I've tried adding another approach. Would this work for you? The RAM thing became too complicated. We can do it another time if needed

Baekalfen:
The advantage is, that it trivially works for every MBC

grungi-ankhfire:
For me this works 👍 I think it's kinda trivial to know which bank to switch things in if you could do it with an absolute address. Maybe one thing is that the docstring says "you can change addresses in ROM 0x0000 to 0x8000", which is true if you think of where these banks are loaded, but at the same time the function works on 0x0000 - 0x4000 (+ bank number). It is obvious if you take even a side glance at the function though :D

And for the RAM, I actually gave it some thought, but I think this is so easily done by writing to the ram after or before each tick() that I am not sure if there would be a real use case?

Baekalfen:
Good spotted. I've fixed the docs.

The RAM could change during a frame, so what I imagined would be to block any writes to the address. But I was worried the solutions I came up with would impact performance

PR #121, opened 2020-04-21
Add basic Kirby Dream Land support

MaciekBaron:
Thought I'll add basic support for Kirby Dream Land:

- Ability to start game (skips start screen and intro)
- Tracking health, lives and score

### DIFF
--- a/pyboy/plugins/game_wrapper_kirby_dream_land.py
+++ b/pyboy/plugins/game_wrapper_kirby_dream_land.py
@@ -8,10 +8,7 @@ __pdoc__ = {
 }
 
 import logging
-from math import floor
-from array import array
 
-import numpy as np
 from pyboy.utils import WindowEvent
 
 from .base_plugin import PyBoyGameWrapper
@@ -44,10 +41,10 @@ class GameWrapperKirbyDreamLand(PyBoyGameWrapper):
         """The lives remaining provided by the game"""
         self.fitness = 0
         """
-        A built-in fitness scoring. The scoring is equals to `score`.
+        A built-in fitness scoring. Taking score, health, and lives left into account.
 
         .. math::
-            fitness = score * health * lives_left
+            fitness = score \\cdot health \\cdot lives\\_left
         """
         super().__init__(*args, game_area_section=(0, 0) + self.shape, game_area_wrap_around=True, **kwargs)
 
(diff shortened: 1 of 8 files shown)


Baekalfen:
That's really cool! Awesome work!

Do you have a small example, that I could include in the automatic testing?

MaciekBaron:
Yeah sure, I'll create an example in the `examples` folder, will that do?

Baekalfen:
Yes, that would be perfect

MaciekBaron:
Ok that's added one, I tested it locally and all assertions pass

pyboy/plugins/game_wrapper_kirby_dream_land.py
```diff
@@ -0,0 +1,172 @@
...
+        self._tile_cache_invalid = True
+        self._sprite_cache_invalid = True
+
+        multiplier = 100000
+        index = 0
+        score = 0
+        while multiplier > 10:
+          score += self.pyboy.get_memory_value(0xD06F + index) * multiplier
+          multiplier /= 10
+          index += 1
```
Baekalfen, line 64: I can't tell what this part is doing

MaciekBaron: The score is stored between `0xd06F`and five bytes later, with every digit of the score taking one byte (later used for displaying the score on the screen).

So if the score is `007250`, I multiply each digit (`0`, `0`, `7`, `2` etc.) by the multiplier (`100000`, `10000`, `1000`, `100` etc.) to get the score value.

Hope this makes sense - there might be a better way of achieving this - I think you already employ a similar approach when reading the score from the screen and using the tile indexes.

Baekalfen: Ah, of course! At a glance, I couldn't figure out what was going to be exponential like that.

Baekalfen:
I fixed a few minor things. I'm no AI/ML expert, but is it important to have the level progressing (how far to the right the AI gets) as a component of the `fitness` value?

If not, I think I'm ready to merge.

And just a question. How did you locate those memory addresses?

MaciekBaron:
The fitness value is currently just a "placeholder", can be improved later on.
I got the memory values using [BGB](https://bgb.bircd.org/) (an emulator with memory viewing/editing capabilities).

Baekalfen:
Ok, understood. But thanks a lot for your contribution! It's pretty solid work :) 

I'll merge it into a pre-release branch now and hopefully soon deploy it as v1.0.2.

PR #125, opened 2020-04-24
Added get_input method to pyboy.py + Added str and int conversion to WindowEvent

Kugge:
I don't know if the documentation for the get_input method is good thought I'm not a great english speaker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 __If you have any questions, or just want to chat, [join us on Discord](https://discord.gg/Zrf2nyH).__
 
 
-It is highly recommended to read the [report](https://github.com/Baekalfen/PyBoy/raw/master/PyBoy.pdf) to get a light introduction to Game Boy emulation. The report is relevant, eventhough you want to contribute to another emulator, or create your own.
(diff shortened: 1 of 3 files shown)


Baekalfen:
There's a problem with using `handle_events`. Each plugin will perform an action based on the input that is given. If you call `pyboy.tick()` and then `handle_event`, the list will likely be empty, and if it isn't, some plugins might do the same operation twice.

I'll see if I can make a way to get the events of the previous cycle.

Kugge:
> There's a problem with using `handle_events`. Each plugin will perform an action based on the input that is given. If you call `pyboy.tick()` and then `handle_event`, the list will likely be empty, and if it isn't, some plugins might do the same operation twice.
> 
> I'll see if I can make a way to get the events of the previous cycle.

Is there a way to convert `self.mb.interaction.standard` and `self.mb.interaction.directional` to `WindowEvent` input ? It should solve the problem i guess

Baekalfen:
> Is there a way to convert `self.mb.interaction.standard` and `self.mb.interaction.directional` to `WindowEvent` input ? It should solve the problem i guess

I'm not sure I follow. I've committed a fix, and it should work now.

PR #134, opened 2020-04-28
Implemented v1 of the Debugger's Memory Window

kr1tzy:
- Implemented the `MemoryWindow` class in pyboy/plugins/debug.py
- Added the necessary attributes in pyboy/utils.py & pyboy/plugins/window_sdl2.py
- Added the new memory window hot keys to the __main__.py output
  - J - next 0x100 bytes
  - K - back 0x100 bytes
  - Left Shift - forward 0x1000 bytes
  - Right Shift - back 0x1000 bytes
- Added pyboy/plugins/assets/Mono_Hack_Font.ttf as the default font to use with SDL2_ttf
- Updated the README with the package manager install instructions for each OS.


kr1tzy:
Forgot to add u guys as  reviewers @Baekalfen @krs013

README.md
```diff
@@ -54,9 +54,10 @@ Original Developers
  * Mads Ynddal - [baekalfen](https://github.com/Baekalfen)
  * Troels Ynddal - [troelsy](https://github.com/troelsy)
 
-GitHub Collaborators
---------------------
+Core Developers
+---------------
 
+ * Mads Ynddal - [baekalfen](https://github.com/Baekalfen)
  * Kristian Sims - [krs013](https://github.com/krs013)
```
Baekalfen, line 61: I removed your name here, as I can see that it was ambiguous.

We'll add a different list for contributors in v1.1.0

pyboy/plugins/debug.py
```diff
@@ -8,11 +8,13 @@
 from array import array
 
 import sdl2
+from sdl2.sdlttf import *
```
Baekalfen, line 11: Could you expand the `*` to be the explicit imports? It makes it easier to work with in the future.

pyboy/utils.py
```diff
@@ -151,7 +151,11 @@ class WindowEvent:
         _INTERNAL_RENDERER_FLUSH,
         _INTERNAL_MOUSE,
         _INTERNAL_MARK_TILE,
-    ) = range(36)
+        PRESS_BUTTON_J,
+        PRESS_BUTTON_K,
+        PRESS_BUTTON_LSHIFT,
+        PRESS_BUTTON_RSHIFT
```
Baekalfen, line 157: The events should represent the action they are performing, not the button

pyboy/plugins/debug.py
```diff
@@ -522,6 +539,119 @@ def update_title(self):
...
+        return events
+
+    def draw_overlay(self):
+        pass
+
+    def copy_tile(self, tile_cache0, t, xx, yy, to_buffer):
+        pass
+
+    def mark_tile(self, x, y, color, height, width, grid):
+        pass
```
Baekalfen, line 652: Did you need to define these?

Baekalfen:
The rendering looks a bit rough. Is there anything that can be done to that? 
image

pyboy/utils.py
kr1tzy: Okay, i can change this.

pyboy/plugins/debug.py
kr1tzy: I believe i did that to just override the default behavior with a pass so it wouldn't do anything wonky.

kr1tzy:
I can mess with the rendering to polish it up some. I'm not fully clear on what you mean by anti-aliasing.

kr1tzy:
Also, I'll take a swing at implementing the Cython stuff. @Baekalfen

Baekalfen:

krs013:
The commit history of this PR has gotten quite polluted, possibly from rebases or force pushes on the upstream branch. In an effort to fix this... I made a copy of this branch and rebased it as well. Maybe not the great solution, but it has a clean history of only the pertinent commits. I haven't force pushed it onto this branch yet; we can decide whether to open a new PR or overwrite this one together. For now, you can see it here: https://github.com/krs013/PyBoy/tree/memory-window-rebase

For my part, I like this. It seems to work well as far as I can tell. I might ask for a different button for fast scrolling than the two shift buttons, though. We've also talked about whether it's worth adding `sdl2-ttf` as a dependency, and I don't remember where we arrived on that. Personally I am against adding dependencies for the main code when possible, since it means existing installations will break with an update. To avoid that, we could try to make the memory window optional somehow (but that would probably mean changing the compilation process), or go for a solution that doesn't use the ttf library (but that sounded like it would require bundling a bitmap font), or wait until some future date when we can link externally built plugins and push this to a separate repo, and reject merging for now (but who knows if that will happen). So... those are the options I see, or else we just add ttf. Either way, this PR has been sitting for a while, so let's see what we can do with it!

krs013:
(Copy of my comment when I did this on the other PR)

I've reset the base branch and cherry picked the commits that belong in this PR. Unfortunately, since that technically creates new commits, the time stamps have changed, so the comments are a little out of order. Still, we can at least see the real history and file changes now. Sorry about the force pushes. I know it's a hassle, but hopefully we don't have to do this too much.

If anyone wants to keep working on this branch, I'd recommend just deleting your local branch and checking out this one. If you have uncommitted changes, you can stash them before the reset and pop them after to reapply. If you have committed changes you can rebase or cherry pick them. Merge will not work anymore, though. This may not be the best solution, but it's what I've got.

krs013:
Anyway, awesome, this is a good feature that we should be able to prepare and merge without too much hassle. For now, I think it doesn't work on PyPy, and we'll want to undo some of the stuff adding TTF and find a way to get non-TTF fonts bundled in. If possible, I think it would be really nice to make the text stuff a general utility that could even be used to decorate some of the other windows, or display status messages/menus on the main screen. Then the memory window display just has to format its contents and call that general print function. I might play around with that a bit, and of course we'll all be talking on discord about how that could go.

pyboy/plugins/debug.py
```diff
@@ -4,22 +4,26 @@
...
 
-
 try:
-    from cython import compiled
+    from cython import compiled, NULL
+    import cython
     cythonmode = compiled
 except ImportError:
     cythonmode = False
+    exec("NULL = None", globals(), locals())
```
krs013, line 25: This is a really clever trick, and I love it. We might be able to work that into some of the other SDL calls.

krs013:
The crashes I mentioned in the last commit message seem to have been a problem with my machine, which had other stability problems as well. I haven't been able to reproduce any crashes on other systems, including another machine with the same software installed and the same machine with a new OS installed. Speed is still a bit of an issue and might deserve some attention, if for nothing else than to be able to independently disable this plugin apart from the existing debug plugins. Otherwise, I think it's ready for review and merge.

Also FWIW GitHub likes to point out that the branch is out-of-date with master, but I don't think we necessarily want to merge master into here. I would locally merge this branch into master directly on a live machine, fixing conflicts and running tests, and then just push the merged master if it all looks good. But maybe I'm missing some idea there, idk.

PR #136, opened 2020-04-29
Feature/gymboy

MathisFederico:
FIxed tests


pyboy/plugins/game_wrapper_tetris.py
```diff
@@ -175,6 +175,7 @@ def game_area(self):
         """
         return PyBoyGameWrapper.game_area(self)
 
+
```
Baekalfen, line 178: Try installing this: `pip install pre-commit`. It will format the code on `git commit` according to the project settings.

PR #144, opened 2020-04-30
Update of mario game_wrapper

MathisFederico:
Added the game_over function
Started the compressed observation for world 1 (assuming the tiles are not the same for every world, we need a better implementation or tiles_compressed for that ...)

Work In Progress


pyboy/plugins/game_wrapper_tetris.py
```diff
@@ -134,9 +147,6 @@ def reset_game(self):
         else:
             logger.error("Tried to reset game, but it hasn't been started yet!")
 
-    def _game_area_np(self):
-        return np.asarray(self.game_area())
-
```
Baekalfen, line 139: This is needed for `__repr__`

pyboy/plugins/game_wrapper_super_mario_land.py
```diff
@@ -21,6 +22,140 @@
 except ImportError:
     cythonmode = False
 
+tiles_minimal = np.zeros((24, 16), dtype=np.uint8)
```
Baekalfen, line 25: I think a lot of the assignments below could be parameterized/automated. There's only hardcoded values. Would it improve readability as well, if it was structure as a for-loop and the ids were automatically incremented?

pyboy/plugins/game_wrapper_super_mario_land.py
```diff
@@ -166,6 +303,13 @@ def game_area(self):
         """
         return PyBoyGameWrapper.game_area(self)
 
+    def game_over(self):
+        if self.lives_left > 0:
+            return False
+        tiles = self._game_area_np()
+        mario_at_bottom = np.any(np_in_mario_tiles(tiles[-1, :]))
+        return np.any(tiles == 15) or mario_at_bottom
```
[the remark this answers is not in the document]
MathisFederico: It's not that minor given that done is run every frame !
It needs to be as clean as it can be, thank you for the advice !

pyboy/plugins/game_wrapper_super_mario_land.py
MathisFederico: Yeah, I know it was a first shot, and I wanted feedback on the tiles_compressed and tiles_minimal before investing some more times !

I'll try to think to a cleaver way now ^^

MathisFederico: Done

Baekalfen:
We might need a test to verify `game_over`

pyboy/plugins/game_wrapper_super_mario_land.py
```diff
@@ -166,6 +235,13 @@ def game_area(self):
         """
         return PyBoyGameWrapper.game_area(self)
 
+    def game_over(self):
+        if self.lives_left > 0:
+            return False
+        tiles = self._game_area_np()
+        mario_at_bottom = np.any(np_in_mario_tiles(tiles[-1, :]))
+        return mario_at_bottom or np.any(tiles == 15)
```
Baekalfen, line 243: We might need a test to verify game_over

pyboy/plugins/game_wrapper_super_mario_land.py
```diff
@@ -30,6 +97,8 @@ class GameWrapperSuperMarioLand(PyBoyGameWrapper):
     If you call `print` on an instance of this object, it will show an overview of everything this object provides.
     """
     cartridge_title = "SUPER MARIOLAN"
+    tiles_compressed = tiles_compressed
+    tiles_minimal = tiles_minimal
```
Baekalfen, line 101: Do we have tests covering these?

pyboy/plugins/game_wrapper_super_mario_land.py
MathisFederico: Yeah but for that we need ROM manipulations of some sort, otherwise it will take a lot of time to reach the states to test (and it's hard to hard code).
Can we load a saved state for tests ?

pyboy/plugins/game_wrapper_super_mario_land.py
MathisFederico: Nope, same that for previous problem

pyboy/plugins/game_wrapper_super_mario_land.py
Baekalfen: I guess it's ok only to test that code is working. We don't need to exhaustively test that every classification is correct.

pyboy/plugins/game_wrapper_super_mario_land.py
Baekalfen: I guess it's ok only to test that code is working. We don't need to exhaustively test that every classification is correct.

pyboy/plugins/game_wrapper_super_mario_land.py
Baekalfen: Maybe we could reduce his "lives"? https://datacrystal.romhacking.net/wiki/Super_Mario_Land:RAM_map

Issue #99, opened 2020-01-02
Load state has corrupted RTC


Baekalfen:
Loading a saved state somehow corrupts the RTC. Maybe if the timestamp in the saved state drifts apart from the real clock time.

Baekalfen:
Possibly solved in #329. Closing until proven otherwise.

PR #147, opened 2020-05-01
Add randomization to game wrappers

### LINKED ISSUE #141: Randomize first Tetromino
When using the game_wrapper, the first Tetromino will always be the same.

Implement a function, which scrambles the registers and provides a new starting Tetromino at every call to `start_game` and `reset_game`.

nicoeps:
Fixes #141 
Add seed and randomize parameters to start_game and reset_game
Add set_seed method to Tetris game wrapper

Does currently not randomize the first three tetrominos after reset_game is called, because the state is saved after the game has already been started. Saving the state earlier fixes this issue, but requires having one more button press to enter the game again.


Baekalfen:
I've tried pushing a different take on it. It generalizes how we could set the timer's DIV register. And adds a specific fix for Tetris, so it won't produce the same sequence in the beginning.

What do you think about this approach?

nicoeps:
The reason for calling it just "seed" was to have a general way of setting values on startup other than just the divider, if some game needed it. _randomize_next_tetromino, however, does bypass Tetris' own generator, which I do not think is an optimal solution. I do not do AI, but I would imagine training with tetrominos that are chosen uniformly at random, instead of with Tetris' own generator, which does not do that, could give different results.

Baekalfen:
WTF, didn't we have save/load state in timer before?! Maybe it's causing this https://github.com/Baekalfen/PyBoy/issues/99. The RTC is a guess, it just causes Pokemon Gold to soft-lock with phone calls from trainers.

pyboy/plugins/game_wrapper_tetris.py
```diff
@@ -110,7 +114,11 @@ def start_game(self):
                 break
 
         # Start game. Just press Start when the game allows us.
-        for _ in range(3):
+        for i in range(3):
+            if i == 2:
+                PyBoyGameWrapper._set_timer_div(self, timer_div, randomize_div)
```
Baekalfen, line 119: This could be called earlier. For example from `PyBoyGameWrapper.start_game(self, timer_div=timer_div, randomize_div=randomize_div)`, as that's what `reset_game` does.

It just need to be set at some point before `save_state`.

pyboy/plugins/game_wrapper_tetris.py
```diff
@@ -120,19 +128,22 @@ def start_game(self):
...
-            self.saved_state.seek(0)
-            self.pyboy.load_state(self.saved_state)
-            self.post_tick()
-        else:
-            logger.error("Tried to reset game, but it hasn't been started yet!")
+        PyBoyGameWrapper.reset_game(self, timer_div=timer_div, randomize_div=randomize_div)
+
+        self.pyboy.send_input(WindowEvent.PRESS_BUTTON_START)
+        self.pyboy.tick()
+        self.pyboy.send_input(WindowEvent.RELEASE_BUTTON_START)
```
Baekalfen, line 143: Nice solution

pyboy/plugins/base_plugin.py
```diff
@@ -122,21 +123,49 @@ def enabled(self):
...
         raise NotImplementedError("post_tick not implemented in game wrapper")
 
-    def start_game(self):
+    def _set_timer_div(self, timer_div, randomize_div):
+        if timer_div is not None:
+            self.pyboy.mb.timer.DIV = timer_div & 0xFF
+        elif randomize_div:
+            self.pyboy.mb.timer.DIV = random.getrandbits(8)
+
+    def start_game(self, timer_div=None, randomize_div=False):
```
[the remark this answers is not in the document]
nicoeps: Added it so the user had an option of not affecting the divider register at all. Not sure how else to do it, but do feel free to change it if you want to.

pyboy/plugins/game_wrapper_tetris.py
nicoeps: Did it like that so that both start_game and reset_game would give the same exact state when called, as there will be more ticks in start_game. Not sure if this is really needed or not, but thought it would be good to let the user have the option of having the exact same value each time, if they provide a specific value for the divider. It could be moved before the loop, where the state could also be saved. Then have three button presses in reset game instead of one. Not sure how much this will impact performance though. PyBoyGameWrapper.start_game gives a warning if called after frame 0, which is why I did not use it. Do let me know if you have a better idea.

pyboy/plugins/base_plugin.py
```diff
@@ -122,21 +123,49 @@ def enabled(self):
...
         """
-        raise NotImplementedError("start_game not implemented in game wrapper")
 
-    def reset_game(self):
+        if not self.pyboy.frame_count == 0:
+            logger.warning("Calling start_game from an already running game. This might not work.")
+
+        self._set_timer_div(timer_div, randomize_div)
+
+    def reset_game(self, timer_div=None, randomize_div=False):
```
MathisFederico, line 152: Same here
Do you want me to commit to show what I mean ?

pyboy/plugins/base_plugin.py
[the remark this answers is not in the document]
nicoeps: > The arguments are strange, here if we want to turn on the randomizer, we have to call `game_wrapper._set_timer_div(None, True)` which is unusual.
> 
> I would think that this is randomized by default and set to a fixed value if called.
> Also, it is more common to have seed in the name like `_set_seed`.

Method arguments have been updated. timer_div was preferred over using seed.

pyboy/plugins/game_wrapper_tetris.py
```diff
@@ -150,7 +152,7 @@ def reset_game(self, timer_div=None):
         """
         After calling `start_game`, you can call this method at any time to reset the game.
 
-        Args:
+        Kwargs:
```
MathisFederico, line 155: I might have to pass the kwargs of openai_gym to reset_game too then

Baekalfen: Should that be done before merging?

MathisFederico: Yeah maybe, we need to test that this works well also (maybe on tetris with a fixed seed)

Baekalfen: Would you do that?

MathisFederico: I don't have the time this week sadly :'(

PR #159, opened 2020-05-04
Optimize bootrom and add sound effect

Powerlated:
I have reduced the size of the core code, freeing up enough space to add a sound effect.
Also moved enabling the LCD to after all graphics are loaded, to prevent Mode 3 from interfering with loading the data in.

Debugged on BGB, can be safely assumed that all issues are with PyBoy, not the bootrom itself.

### DIFF
--- a/pyboy/plugins/window_sdl2.pxd
+++ b/pyboy/plugins/window_sdl2.pxd
@@ -12,8 +12,9 @@ cimport cython
 from libc.stdint cimport uint8_t, uint16_t, int16_t, uint32_t
 
 
-
 cdef int ROWS, COLS
+
+cdef sdl2.SDL_GameController *_sdlcontroller = NULL
 
 cpdef list sdl2_event_pump(list)
 
--- a/pyboy/plugins/window_sdl2.py
+++ b/pyboy/plugins/window_sdl2.py
@@ -55,10 +55,36 @@ KEY_UP = {
     sdl2.SDLK_COMMA     : WindowEvent.RELEASE_REWIND_BACK,
     sdl2.SDLK_PERIOD    : WindowEvent.RELEASE_REWIND_FORWARD,
 }
+
+CONTROLLER_DOWN = {
+    sdl2.SDL_CONTROLLER_BUTTON_DPAD_UP       : WindowEvent.PRESS_ARROW_UP,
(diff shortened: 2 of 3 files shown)


krs013:
Love the sound! This is awesome.

You'll see I made a couple commits, one to build the result and put it in version control, and another to normalize the indentation. You can redo these and force push on your account if the credit matters, since obviously you've compiled this before. Either way, though, the register names need to be uniform. I don't know about Baekalfen, but if you want to make the literals and/or registers lowercase, that's fine with me, but register names like `A` or `sp` should be consistently capitalized or lowercased. If you asked me I'd say caps, if you're undecided, but since you lowercased the literals I assume you have a preference.

Also, obviously, a lot of tests are failing now. Some of it may be regression tests that were influenced by the old boot rom, but I'm not up to speed on all the tests so we'll have to ask Baekalfen which of those are okay to reset. For what it's worth, the few games I ran them on were okay, but we added those arbitrary values because some specific ones weren't working, and we thought there was some secret initialization in the real bios rom and just copied the result.

krs013:
Oh, also, a bit unrelated, but apparently the old bottom doesn't compile now? Not sure when that happened. Recompiling the bottom may need to become part of the build or test process, or perhaps it can be split out to a different repo or something.

Baekalfen:
I haven't looked at the rest of the PR, but looks very interesting!

I just wanted to say, that the bootrom compiles fine on my machine.

Baekalfen:
I agree with @krs013 about the naming and style. Please use whatever code style is currently in use, and raise it as a separate issue, if the style is flawed.

bootrom/bootrom.asm
```diff
@@ -1,34 +1,43 @@
...
-    ; Set color palette to 11111100
-    ld A, $fc
-    ld [$FF00+$47], A
+    ; Erase VRAM - it is filled with garbage on startup on hardware
+    xor A
+    ld HL,$9fff
+Erase:
+    ld [HL-],A
+    bit 7,H
+    jr nz, Erase
```
Baekalfen, line 11: I can recognize, that this is from the official DMG bootrom. We cannot include anything in the project, that is obviously copied like that. Both so we don't have any case for DMCA notices, but also morally, that we don't copy-paste code from any source. No matter how trivial it is.

Baekalfen:

PR #172, opened 2020-05-25
Create basic_bot_functions.py

DavidSteinhart:
Added example to show basic bot functions usage.

### DIFF
--- a/examples/basic_bot_functions.py
+++ b/examples/basic_bot_functions.py
@@ -1,0 +1,104 @@
+from pyboy.utils import WindowEvent
+import pyboy as pyboy
+import io
+
+'''
+This shows the basic functions you can use if you want to write a bot using only the game image as a human would see it.
+Uses Super Mario Land 1 as an example.
+'''
+
+path_to_rom = '../SuperMarioLand1.gb'
+
+
+def save_state_to_file(boy, path):
+	boy.save_state(open(path, "wb"))
+
+
+def load_state_from_file(boy, path, reset_keys=False):
+	boy.load_state(open(path, "rb"))
+	if reset_keys:
+		for key in range(WindowEvent.RELEASE_ARROW_UP, WindowEvent.RELEASE_BUTTON_START + 1):
+			boy.send_input(key)
+
+
+def save_state(boy):
+	file_like_object = io.BytesIO()
+	file_like_object.seek(0)
+	boy.save_state(file_like_object)
+	return file_like_object
+
+
+def load_state(boy, file_like_object, reset_keys=False):
+	file_like_object.seek(0)
+	boy.load_state(file_like_object)
+	if reset_keys:
+		for key in range(WindowEvent.RELEASE_ARROW_UP, WindowEvent.RELEASE_BUTTON_START + 1):
+			boy.send_input(key)
+
+
+def show_screen(screen):
+	img = screen.screen_image()
+	img.show()
+
+
+def save_screen(screen, path):
+	img = screen.screen_image()
+	img.save(path)
+
+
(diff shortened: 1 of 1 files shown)


examples/basic_bot_functions.py
```diff
@@ -0,0 +1,104 @@
+from pyboy.utils import WindowEvent
+import pyboy as pyboy
+import io
+
+'''
+This shows the basic functions you can use if you want to write a bot using only the game image as a human would see it.
+Uses Super Mario Land 1 as an example.
+'''
+
+path_to_rom = '../SuperMarioLand1.gb'
```
Baekalfen, line 10: Have a look in the other examples. It would be better if the user can specify the location

examples/basic_bot_functions.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def save_state_to_file(boy, path):
+	boy.save_state(open(path, "wb"))
```
Baekalfen, line 14: You should probably use the context manager `with open(...) as f:`

examples/basic_bot_functions.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def load_state_from_file(boy, path, reset_keys=False):
+	boy.load_state(open(path, "rb"))
+	if reset_keys:
+		for key in range(WindowEvent.RELEASE_ARROW_UP, WindowEvent.RELEASE_BUTTON_START + 1):
+			boy.send_input(key)
```
Baekalfen, line 21: You're indenting with 6 spaces. The project uses 4 spaces.

examples/basic_bot_functions.py
Baekalfen, line 10: You should install `pip install pre-commit`. This will take care of formating according to the project's settings.

examples/basic_bot_functions.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def save_state(boy):
+	file_like_object = io.BytesIO()
+	file_like_object.seek(0)
+	boy.save_state(file_like_object)
```
Baekalfen, line 27: Did you mean to flip the order of `seek` and `save`?

examples/basic_bot_functions.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def save_state_to_file(boy, path):
+	boy.save_state(open(path, "wb"))
+
+
+def load_state_from_file(boy, path, reset_keys=False):
+	boy.load_state(open(path, "rb"))
+	if reset_keys:
+		for key in range(WindowEvent.RELEASE_ARROW_UP, WindowEvent.RELEASE_BUTTON_START + 1):
```
Baekalfen, line 20: Arithmetic on WindowEvents is not encouraged. They are numbers at the moment, but that is not a guarantee. It is also not clear, which event you mean to use.

examples/basic_bot_functions.py
DavidSteinhart: I was unsure about that as well. The only thing I found in your documentation was the following:
# Save to memory
import io
file_like_object = io.BytesIO()
file_like_object.seek(0)

so I just copied that. That's why I created this file, because I could not find any examples of how to use these functions properly.

examples/basic_bot_functions.py
DavidSteinhart: Sorry about that, I installed pre-commit but did not check if it worked.

DavidSteinhart:
So should I close the pull request?

Baekalfen:
Sorry, I think we should. But I would like to again thank you for your interest and contribution! I do appreciate it!

PR #178, opened 2020-07-30
Debugger functionality and plugin

Baekalfen:


pyboy/core/cpu.py
```diff
@@ -168,35 +89,107 @@ def load_state(self, f, state_version):
...
-            f"State loaded: A:{self.A:02x}, F:{self.F:02x}, B:{self.B:02x}, C:{self.C:02x}, D:{self.D:02x}, E:{self.E:02x}, HL:{self.HL:02x}, SP:{self.SP:02x}, PC:{self.PC:02x}, IME:{self.interrupt_master_enable}, halted:{self.halted}, stopped:{self.stopped}"
+        if state_version >= 5:
+            # Interrupt register moved from RAM to CPU
+            self.interrupts_enabled_register = f.read()
+        logger.debug("State loaded: " + self.dump_state())
+
+    def dump_state(self):
+        return (
+            f"A:{self.A:02x}, F:{self.F:02x}, B:{self.B:02x}, C:{self.C:02x}, D:{self.D:02x}, E:{self.E:02x}, "
+            f"HL:{self.HL:02x}, SP:{self.SP:02x}, PC:{self.PC:02x}, "
```
krs013, line 100: PC, SP, and HL probably should be formatted with 04x

pyboy/core/cpu.py
```diff
@@ -168,35 +89,107 @@ def load_state(self, f, state_version):
...
+
+    def tick(self):
+        if self.check_interrupts():
+            self.halted = False
+            # TODO: We return with the cycles it took to handle the interrupt
+            return 0
+
+        if self.halted and self.interrupt_queued:
+            # GBCPUman.pdf page 20
+            # WARNING: The instruction immediately following the HALT instruction is "skipped" when interrupts are
```
krs013, line 120: Is this implemented here, or just a note? It doesn't seem like it emulates the bug, but hopefully that's not too important to roms that properly avoid it.

pyboy/core/cpu.py
```diff
@@ -168,35 +89,107 @@ def load_state(self, f, state_version):
...
+            # disabled (DI) on the GB,GBP, and SGB.
+            self.halted = False
+            self.PC += 1
+            self.PC &= 0xFFFF
+        elif self.halted:
+            return 4 # TODO: Number of cycles for a HALT in effect?
+
+        old_pc = self.PC
+        cycles = self.fetch_and_execute(self.PC)
```
krs013, line 129: This isn't necessarily new but I just noticed that we're passing a reference to `self.PC` and incrementing the value by reference in `fetch_and_execute(...)`. Is there a reason for this, or could we remove the `pc` argument from `fetch_and_execute` and have it use `self.PC` directly? That would be more clear and less likely to turn into a bug in the future.

pyboy/core/cpu.py
```diff
@@ -168,35 +89,107 @@ def load_state(self, f, state_version):
...
+            self.is_stuck = True
+        self.interrupt_queued = False
+        return cycles
+
+    def check_interrupts(self):
+        if self.interrupt_queued:
+            # Interrupt already queued. This happens only when using a debugger.
+            return False
+
+        if (self.interrupts_flag_register & 0b11111) & (self.interrupts_enabled_register & 0b11111):
```
krs013, line 141:
```suggestion
        if (self.interrupts_flag_register & self.interrupts_enabled_register & 0b11111):
```

pyboy/core/cpu.py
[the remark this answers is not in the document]
Baekalfen: Done. Also merged it with the code in `debug.py`

pyboy/core/cpu.py
Baekalfen: That's just a quote from the manual. It was actually there already. It's not implemented.

Baekalfen: Also, I have never heard of a practical example that depends on this.

pyboy/core/cpu.py
Baekalfen: We should probably remove the argument. But we are incrementing only the local variable in the function.

pyboy/core/cpu.py
Baekalfen: I can see what you want to do, but I'm worried someone (me) won't be able to immediately recognize the semantics of this, and might need to be explained with a comment. And the compiler will optimize it anyway.

krs013: Yeah no worries; I'll drop this one

pyboy/core/cpu.py
krs013: Guess I didn't see it in the old source. I made that comment before I fully understood how the PC was being manipulated (and I still might not get it, based on the other one) so I thought the `+=1` was for the bug.

Baekalfen: I think the +1 you're look at is just to escape the HALT instruction

pyboy/core/lcd.pxd
```diff
@@ -3,12 +3,15 @@
...
 from cpython.array cimport array
-from libc.stdint cimport uint8_t, uint16_t, uint32_t, uint64_t
+from libc.stdint cimport uint8_t, uint16_t, uint32_t, uint64_t, int16_t
 cimport pyboy.utils
+from pyboy.utils cimport color_code
 from pyboy.utils cimport IntIOInterface
-
+cdef uint8_t INTR_VBLANK, INTR_LCDC, INTR_TIMER, INTR_SERIAL, INTR_HIGHTOLOW
 cdef uint16_t LCDC, STAT, SCY, SCX, LY, LYC, DMA, BGP, OBP0, OBP1, WY, WX
 cdef int ROWS, COLS, TILES, VIDEO_RAM, OBJECT_ATTRIBUTE_MEMORY
```
krs013, line 16:
```suggestion
cdef int ROWS, COLS, TILES, FRAME_CYCLES, VIDEO_RAM, OBJECT_ATTRIBUTE_MEMORY
```

setup.py
```diff
@@ -258,7 +274,7 @@ def get_export_symbols(self, ext):
     ext_modules = cythonize(
         [*cythonize_files], # This runs even if build_ext isn't invoked...
         nthreads=thread_count,
-        annotate=False,
+        annotate=True,
```
krs013, line 277: Did you mean to commit this? Fine if so, just checking.

krs013 (dismissed):
Sorry, that took a little while. This is a really cool feature! And it brings a lot of good changes and fixes too. I'm excited to see it done.

setup.py
Baekalfen: You're right. That was a mistake

Baekalfen:
@

> Sorry, that took a little while. This is a really cool feature! And it brings a lot of good changes and fixes too. I'm excited to see it done.

Thanks for the review!

PR #179, opened 2020-09-23
Sound control

krs013:
Added support for the sound control registers. The sound module is still experimental/beta, but this fixes some of the incorrect sounds we were experiencing before.


Baekalfen (dismissed):
Looks great! I'm trying to test it locally, and fix Travis. Then we could be sending this straight to PyPi.

PR #182, opened 2020-10-01
Add controller support

nicoeps:
Add controller support because why not

### DIFF
--- a/pyboy/plugins/window_sdl2.pxd
+++ b/pyboy/plugins/window_sdl2.pxd
@@ -12,8 +12,9 @@ cimport cython
 from libc.stdint cimport uint8_t, uint16_t, int16_t, uint32_t
 
 
-
 cdef int ROWS, COLS
+
+cdef sdl2.SDL_GameController *_sdlcontroller = NULL
 
 cpdef list sdl2_event_pump(list)
 
--- a/pyboy/plugins/window_sdl2.py
+++ b/pyboy/plugins/window_sdl2.py
@@ -83,8 +83,6 @@ CONTROLLER_UP = {
 }
 # yapf: enable
 
-_sdlcontroller = None
-
 def sdl2_event_pump(events):
     global _sdlcontroller
     # Feed events into the loop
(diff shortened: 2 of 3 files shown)


krs013:
Like I said on discord, this is an awesome feature that I've wanted for a while, and it works great on my machine so far. Haven't tested on anything besides linux but I can later. We may want to discuss a better place for `_sdlcontroller` than as a global variable but otherwise I think this is a pretty easy one to accept!

(Also I haven't looked for myself yet but will this work with `pyboy.PyBoy.get_input()`?)

pyboy/plugins/window_sdl2.pxd
```diff
@@ -12,9 +12,10 @@ cimport cython
 from libc.stdint cimport uint8_t, uint16_t, int16_t, uint32_t
 
 
-
 cdef int ROWS, COLS
 
+cdef sdl2.SDL_GameController *_sdlcontroller = NULL
```
Baekalfen, line 17: Could we move this into the `WindowSDL2` class? Then I'm ready for merging this

Baekalfen: Ok, I see the problem. We'll just merge it as is

krs013:
Before we merge I'd like to quickly test how it works on macOS and Windows. Also, are we happy with the button mapping? I don't have any objections, just mentioning.

Baekalfen:
> Before we merge I'd like to quickly test how it works on macOS and Windows. Also, are we happy with the button mapping? I don't have any objections, just mentioning.

You're welcome to try it out, I'll wait a bit. I've looked at the mapping again. I don't have much of an opinion, but it looks reasonable?

krs013:
I can't get the controller to work on macOS. There's a lot that could be going wrong but I have confirmed that the controller works in at least one other game. Has anyone else had success with macOS?

Linux works perfectly, and Windows works, but it swaps the B/A keys because I have a Wii U controller on an Xinput profile. We might want format switching in the future but I'm fine leaving that out of the scope of this PR. Also, using the home button for speed up activated Steam Big Picture and crashed PyBoy, but after I quit Steam it was fine. So I'm not sure if we need to change that or just punt it to configuration too.

If anyone can confirm it works on macOS then I'm fine to merge.

Baekalfen:
My wireless Xbox 360 controller doesn't work on Mac, so I can't test it. But I'm willing to merge it as-is, as any issues on Mac, likely would be because of SDL, and not PyBoy.

PR #184, opened 2020-10-03
Optimize bootrom and add sound effect

Baekalfen:
Recreated PR #159 from @Powerlated, as GitHub won't let it be reopened.


krs013 (approved):
Looks good to me! I'm really happy with the sound it makes!

PR #190, opened 2021-01-26
Fixing setup.py for Python 3.8/3.9

Baekalfen:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # Change in setup.py
 cython>=0.29.16; platform_python_implementation == 'CPython'
 numpy
-pillow
+pillow>=8.0.0
 pysdl2
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,5 @@
 import distutils.cmd
+import multiprocessing
 import os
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 # Change in setup.py
 cython>=0.29.16; platform_python_implementation == 'CPython'
 numpy
-pillow
+pillow>=8.0.0
```
krs013, line 4: What's the reason for this? It's preventing me from running tests on Python 3.7 on macOS, though I'm not exactly sure why...

krs013: Okay, I seem to be able to work around it, so I guess it's just a quirk with using `python setup.py test`. Tests are running and I haven't had issues on any other platforms.

krs013: Ah, I figured out what the problem was. I had `gym` installed, which requires `Pillow<=7.2.0`. So this probably shouldn't be merged until we address that.

krs013 (changes requested):
Need to fix conflicting requirements with `gym`

PR #192, opened 2021-02-13
Cpu breakpoints

rweisman:
This PR adds breakpoint support to the PyBoy API as client-defined callbacks.  I have a use case for this functionality and thought it might be useful to others as well.

PR is WIP - Looking for feedback as to whether this is appropriate for the project before adding tests etc.  I see there's an ongoing project to add a debugger which would probably supersede this entirely.

Please let me know what you think!

Sample Usage (client code):
```
def on_breakpoint(addr, rambank, rombank):
    print('BREAKPOINT: {} | {} | {}'.format(hex(addr), hex(rambank), hex(rombank)))
    if rombank is 0x1E:
        pyboy.set_memory_value(0xD368, 0xFF)  # Do something useful...


pyboy = PyBoy(gamerom_file=...)
pyboy.set_breakpoint(0x4163, on_breakpoint)

while not pyboy.tick():
    pass
```

### DIFF
--- a/pyboy/core/cpu.pxd
+++ b/pyboy/core/cpu.pxd
@@ -78,3 +78,5 @@ cdef class CPU:
     @cython.locals(v=cython.int)
     cdef bint test_ramregisterflag(self, int, int)
     cdef void clear_ramregisterflag(self, int, int)
+
+    cdef dict breakpoints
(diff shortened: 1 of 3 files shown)


Baekalfen:
Thanks for your PR! I'm actually working on a similar functionality on the `debugger` branch. You should go and check that out. There's no place to attach a callback function, but that could be an option

rweisman:
Thanks for your reply!  Will definitely check out the `debugger` branch.  Closing this PR.

PR #193, opened 2021-03-02
Color Support

Baekalfen:


pyboy/core/cartridge/base_mbc.py
```diff
@@ -37,6 +38,8 @@ def __init__(self, filename, rombanks, external_ram_count, carttype, sram, batte
         self.rambank_selected = 0
         self.rombank_selected = 1
 
+        self.cgb = self.getitem(0x0143) >> 7
```
krs013, line 41:
```suggestion
        self.cgb = bool(self.getitem(0x0143) >> 7)
```
Cython will auto promote this to a boolean because it's a `bint` type but Python and PyPy will think it's an int. Might cause a bug later if we forget

Issue #187, opened 2020-12-04
game_wrapper flag on PyBoy constructor missing


Lafitae:
Hello, as discussed on Discord, while trying to load a gb rom that has an already existing game wrapper implementation, calling pyboy.game_wrapper() returned none and the openaigym would not work. We figured out that this is because game_wrapper=True needs to be passed in the PyBoy constructor, which I had not been able to find anywhere when reading the documentation or source code, so a flag seems to be missing for this?

![image](https://user-images.githubusercontent.com/43269965/101176921-7648ea00-3647-11eb-9048-323ed824c966.png)

![image](https://user-images.githubusercontent.com/43269965/101177069-a98b7900-3647-11eb-8510-124036bc2801.png)

Issue #170, opened 2020-05-24
Send_Input information is not stored when saving / loading state


DavidSteinhart:
When loading a save state, key presses are not stored or reset, so when loading a state and performing input, you will get different results depending on the inputs that were performed before. 

Example:

```
from pyboy.utils import WindowEvent
import pyboy as pyboy
import io


def save_state(boy):
	file_like_object = io.BytesIO()
	file_like_object.seek(0)
	boy.save_state(file_like_object)
	return file_like_object


def load_state(boy, file_like_object):
	file_like_object.seek(0)
	boy.load_state(file_like_object)

# Load Super Mario Land and progress to start of game
boy = pyboy.PyBoy('../SuperMarioLand1.gb')
for i in range(100):
	boy.tick()
boy.send_input(WindowEvent.PRESS_BUTTON_START)
boy.tick()
boy.send_input(WindowEvent.RELEASE_BUTTON_START)
for i in range(20):
	boy.tick()

# Save state of the emulator
state = save_state(boy)

# Send input and progress game
boy.send_input(WindowEvent.PRESS_ARROW_RIGHT)
for i in range(30):
	boy.tick()

# Load state and progress game
load_state(boy, state)
for i in range(30):
	boy.tick()
# Mario keeps running because PRESS_ARROW_RIGHT is still active but it would be expected that he does not move.
```

```
# Use this to reset all input as a quick-fix
for key in range(WindowEvent.RELEASE_ARROW_UP, WindowEvent.RELEASE_BUTTON_START + 1):
	boy.send_input(key)
```

Baekalfen:
Nice find. I guess `interaction.py` might not be covered by the save/load state at all. Although it should of course.

R0cker77:
Hi, i ran the code sample you gave, and mario stopped after the code stopped excuting, so i dont understand what is the issue here, is it the state itself not stopping? appreciate any help..

krs013:
Hi @R0cker77, sorry we never followed up with you on this; I only finally got a good look at this issue now. When I ran the sample code, Mario moved once, reset to his original position, and moved again, which is the wrong behavior. If that wasn't the case when you tried it, I'm not sure why that would be. I think my fix for this should be enough to take care of it, so there's no need to worry about it any more, and thanks for your interest in helping!

PR #196, opened 2021-03-26
Small fixes

krs013:
Here are a few bugfixes, including that font fix I promised, fixes for #170 and #187, and fixes for a couple runtime bugs I encountered on the way.

I also did add the script I used to make the font along with a makefile to guide the process. It should work on Mac/Linux but it only needs to be run to generate the file that's already committed, so it's mostly there if we want to change it or retrace our steps. It's pretty sketchy code, though, please don't judge, haha.

I don't know if there's a sure way for me to test the font/pip situation, but `pip install .` made an installation in a venv that could use the font properly.

There might be some changes needed if the tests fail. I couldn't get them all to pass locally, but I don't think that's always an indicator of how they'll do on Travis.

In all, these should all be pretty simple and small tweaks. Feel free to change anything that doesn't seem right.


pyboy/core/cartridge/cartridge.py
```diff
@@ -57,7 +57,15 @@ def load_romfile(filename):
...
 
     logger.debug(f"Loading ROM file: {len(romdata)} bytes")
+    if len(romdata) == 0:
+        logger.error("ROM file is empty!")
+        raise Exception("Empty ROM file")
+
     banksize = 16 * 1024
+    if len(romdata) % banksize != 0:
+        logger.error("Unexpected ROM file length")
+        raise Exception("Bad ROM file size")
```
Baekalfen, line 67: Exceptions are not working so well in Cython, but this isn't a central part of any important logic, so let's just leave it like this.

pyboy/pyboy.py
```diff
@@ -224,12 +227,16 @@ def stop(self, save=True):
             save (bool): Specify whether to save the game upon stopping. It will always be saved in a file next to the
                 provided game-ROM.
         """
-        if not self.stopped:
+        # If __init__ errors, __del__ will call try to call stop() but
+        # self.stopped doesn't exist yet
+        if hasattr(self, "stopped") and not self.stopped:
```
Baekalfen, line 232: Does `hasattr` work in Cython?

Baekalfen (approved):
Great job! Not a single thing I would have done differently

Baekalfen:
Your history is so neat, that I'll rebase it ;)

Baekalfen:
Great, then it merged anyway...

pyboy/core/cartridge/cartridge.py
krs013: They seemed okay when I tested them, but if it becomes a problem some day we can deal with it. It's not likely that these will fire for anyone--but I made a mistake once that caused a rom to have 0 length, and this is easier to understand than a segfault, haha.

pyboy/pyboy.py
krs013: It seems to? I'm not sure *how* it works, but before I added this there were some weird errors happening on __del__. Maybe `hasattr` returns false if the property isn't initialized or something.

krs013: Update: as we found later, it definitely doesn't work. I pushed a couple more commits that fix it, but I think I'll have to open a separate PR for that. Sorry!

PR #199, opened 2021-04-15
Replace custom scaling logic with SDL_RenderSetLogicalSize

JonathanMurray:
Instead of translating mouse events by keeping track of the
window's scale, let SDL take care of it.

This commit fixes an issue where mouse clicks and hovers would not
point at correct pixels after resizing a window. Also, now aspect
ratio is maintained after resizing.

This PR is a follow-up from https://github.com/Baekalfen/PyBoy/pull/195 which proposed a more complicated and cumbersome way of achieving the same thing.

### DIFF
--- a/pyboy/plugins/debug.py
+++ b/pyboy/plugins/debug.py
@@ -353,6 +353,7 @@ class BaseDebugWindow(PyBoyWindowPlugin):
         self.buf, self.buf0, self.buf_p = make_buffer(width, height)
 
         self._sdlrenderer = sdl2.SDL_CreateRenderer(self._window, -1, sdl2.SDL_RENDERER_ACCELERATED)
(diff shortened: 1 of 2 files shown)


Baekalfen:
This is looking really good! @krs013 Got any comments, or do we just merge?

krs013 (approved):
Looks great to me! I'm fine with merging it; the only change I'd suggest would be to add this to the main game window too so it also gets the letterboxing.

Baekalfen:
> so it also gets the letterboxing.

That's a good idea. Do either of you want to add that? Otherwise, we can just merge

krs013 (approved):

Baekalfen:
Thanks both of you! @JonathanMurray @krs013

krs013:
Congrats on your first PR @JonathanMurray! Thanks for being willing to redo it even after all that work on the first one! Hope you stick around 👍

PR #220, opened 2022-04-18
SDL2 without cython compilation

krs013:
Performance and compatibility seem fine, but needs testing. Also `setup.py` and install requirements should be updated still. Don't merge yet.


pyboy/plugins/debug.py
```diff
@@ -233,7 +237,14 @@ def stop(self):
...
     def enabled(self):
-        return self.pyboy_argv.get("debug")
+        if self.pyboy_argv.get("debug"):
+            if not sdl2:
+                logger.error("Failed to import sdl2, needed for debug window")
+                return False
+            else:
+                return True
+        else:
+            return False
```
Baekalfen, line 247: We can shorten this down. For example:
```python
if not sdl2:
    logger.error("Failed to import sdl2, needed for debug window")
    return False
return self.pyboy_argv.get("debug")
```

pyboy/plugins/window_sdl2.py
```diff
@@ -181,7 +184,14 @@ def post_tick(self):
...
     def enabled(self):
-        return self.pyboy_argv.get("window_type") == "SDL2" or self.pyboy_argv.get("window_type") is None
+        if self.pyboy_argv.get("window_type") in ("SDL2", None):
+            if not sdl2:
+                logger.error("Failed to import sdl2, needed for sdl2 window")
+                return False  # Disable, or raise exception?
+            else:
+                return True
+        else:
+            return False
```
Baekalfen, line 194: Same as before:
```python
if not sdl2:
    logger.error("Failed to import sdl2, needed for SDL2 window")
    return False
return self.pyboy_argv.get("window_type") == "SDL2" or self.pyboy_argv.get("window_type") is None
```
Exceptions are tricky in Cython.

Baekalfen (commented):
Looks good. Haven't tested it yet

pyboy/plugins/debug.py
krs013: Yeah I wasn't thrilled with it, but this function gets called even if the sdl2/debug windows aren't requested, so we don't want to log errors in that case.

pyboy/plugins/window_sdl2.py
krs013:
```suggestion
        if self.pyboy_argv.get("window_type") in ("SDL2", None):
            if sdl2:
                return True
            else:
                logger.error("Failed to import sdl2, needed for sdl2 window")
            return False
```

Not sure there's a shorter way than that if we want it to log the error correctly

Baekalfen:
This has been pulled into #193

PR #233, opened 2022-10-08
Dynamically loading plugins

Baekalfen:


pyboy/plugin_manager.py
```diff
@@ -22,9 +32,14 @@
 enabled_window_plugins = []
 enabled_gamewrappers = []
 
-for mod_name in [x.name for x in iter_modules(plugins.__path__)]:
-    mod = importlib.import_module("pyboy.plugins." + mod_name)
+builtin_plugins = [importlib.import_module("pyboy.plugins." + m.name) for m in iter_modules(plugins.__path__)]
+external_plugins = []
```
Baekalfen, line 36: Probably want something like Django's list of middlewares to manage loading/execution order

PR #234, opened 2022-10-09
Modular Plugin Manager

hexiro:
Allows for modularly adding plugins (game wrappers) to PyBoy. This is helpful so users can create their own game wrappers and add them without them having to be officially supported.

There are still some Cython type definition things missing as that's not my specialty, but besides that, I'd like to hear your thoughts.

P.S. my code formatter accidentally added some spaces between plus signs, so if you have an auto code formatter, you might want to run it on this branch.


hexiro:
I see that coincidentally, you opened a PR with a similar goal 2 hours ago. I just opened this because I never finished it and I was organizing my repos (deleting my forks) so I figured I'd open this. If this doesn't get merged that's okay but I hope it can help serve as a reference for what you're doing :)

Baekalfen:
Thank you! This is actually really helpful. It's interesting now that we have made two different approaches at the same thing. I think I can steal some of your ideas, and improve mine. I'll be sure to give you a ping and set you as co-author on the commit if I do :)

hexiro:
sounds good! glad it can be of use.

Baekalfen:
Closed for inactivity and duplicate of #233

PR #236, opened 2022-10-22
changed cython floats to doubles

krs013:
CPython floats are internally doubles on most modern machines, but in Cython there are distinct float and double types. This PR replaces floats in our `pxd` files to doubles so they match the outputs of the functions they're interacting with.

This also will hopefully fix many of our issues with the frame limiter not working, since the 32-bit floats didn't have enough precision to measure the delay needed.


Baekalfen (approved):
Looks good. If build+test passes, we can merge

PR #237, opened 2022-10-22
Misc. fixes

### LINKED ISSUE #235: requirements.txt installs Cython only on CPython, but setup.py tries to use it on anything that is not PyPy
The requirements txt leads to Cython not being installed on e.g. GraalPy, but then setup.py tries to Cythonize, because GraalPy is not PyPy
https://github.com/Baekalfen/PyBoy/blob/311562aa5737771c0702ec78f6639d1b04f6a0c0/setup.py#L19
https://github.com/Baekalfen/PyBoy/blob/311562aa5737771c0702ec78f6639d1b04f6a0c0/setup.py#L35

Baekalfen:
Fixes #235 as well.


krs013 (dismissed):
Looks good & works on macOS, Windows & Linux from my tests. I'm also going to add some miscellaneous fixes I had that don't change any code.

krs013 (approved):
Guess I should have pushed before reviewing. No functional changes, like I said, so I'll just approve again

PR #243, opened 2023-03-19
fix: typo in state save/load

cmacrae:
saving/loading states was broken :) I think this may have been a typo left behind from debugging.  
thanks for this awesome project!


Baekalfen:
Thank you for your PR! You're right, that was a mistake.

To not break backwards-compatibility, I had to add some extra stuff. If it passes the tests, I will merge it and deploy :)

PR #262, opened 2023-10-25
Compatibility improvements: remove zsh and graceful errors if sound fails

krs013:
What it says in the title. WSL had cryptic float errors if `libpulse0` wasn't installed, now the `sound` module will report an error and revert to emulating the sound.

Also replaced `zsh` in makefile with `find`. This should be available on all Unix systems as far as I know.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,6 @@ PYPY := pypy3
 ROOT_DIR := $(shell git rev-parse --show-toplevel)
 GITHUB_REF ?= "refs/tags/v0.0.0"
 PYBOY_VERSION ?= $(shell echo ${GITHUB_REF} | cut -d'/' -f3)
-SHELL:=zsh -o extended_glob +o nomatch -c
 
 dist: clean build
 	${PY} setup.py sdist bdist_wheel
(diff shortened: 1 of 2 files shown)


Baekalfen:
I rebased your code so we don't have the merge commit, and the extra commit, that fixes logging in the first. Also, I removed the `self.device = 0`, as that isn't need.

krs013:
Sounds good, and sorry for the messy history. I appreciate the changes

`self.device = 0` was just there so the module doesn't leave `__init__()` with that field undeclared, but that's more of a tidiness thing than functional. It might lead to errors that won't show up in Python but will in CPython/PyPy. If there's a better place to put it, that's fine--we could even have it initialized to 0 before calling `SDL_Init` and leave it at that. Not a big deal either way once we remove SDL code from `sound.py`, which is still something I intend to do to match how we did `lcd.py`.

Baekalfen:
> It might lead to errors that won't show up in Python but will in CPython/PyPy

I checked. All uses are protected by `if self.enabled`, so it won't be a problem.

It would be awesome if you have time to move the SDL2 stuff out, like you say.

Baekalfen (approved):

PR #263, opened 2023-10-25
Fix Kirby score variable and game over condition

eugeneyjy:
1) Add one more number of digits (up to 100 000s) and fix the order of score digits from RAM lookup.

2) Workaround and fix game over condition. Due to the way Kirby's Dream Land work, the `game_over()` can't be simply do `return self.lives_left == 0 and self.health == 0`. When Kirby has 1 lives left, in reality, the player can still die twice (having 2 lives). However, when the player die with 1 lives left, their health will remain 0 while their lives left will get set to 0. Although this only happens in a split seconds, it will make the `game_over()` function `return True` even though the player still have 1 more live to go. It seems like `self.lives_left` can't go below 0 as well, that's why I go with this workaround. Not sure if that's the best way or against the practice.

### DIFF
--- a/pyboy/plugins/game_wrapper_kirby_dream_land.pxd
+++ b/pyboy/plugins/game_wrapper_kirby_dream_land.pxd
@@ -14,3 +14,4 @@ cdef class GameWrapperKirbyDreamLand(PyBoyGameWrapper):
     cdef public int health
     cdef public int lives_left
     cdef public int fitness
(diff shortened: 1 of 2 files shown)


Baekalfen:
Looks good to me. I've approved the test runners, so let's see what they say. If all good, let's merge it :)

Baekalfen:
Thanks for the patch!

PR #264, opened 2023-10-26
Update gymnasium

Sisyphus192:
The original OpenAI [gym](https://github.com/openai/gym) is no longer in development and has been succeeded by [gymnasium](https://github.com/Farama-Foundation/Gymnasium).

This PR:
- Updates the requirements to use `gymnasium` instead of `gym`.
- Modifies the code to import from `gymnasium` instead of `gym`. 
- Maintains the naming "openai_gym" in various places for name recognition.

Changes to API Calls:
- `step()` and `reset()` functions, along with relevant tests, have been updated to match the new gymnasium [API](https://gymnasium.farama.org/api/env/#gymnasium.Env.step).
  - Notably, in version 0.26 of gymnasium: The Step API changed, removing `done` in favor of `terminated` and `truncated` to clarify to users when the environment had terminated or truncated, which is crucial for reinforcement learning bootstrapping algorithms.
  - Consequently, language was updated from `done` to `terminated` to align with this API change.

Notes:
- The default returned value for `truncated` is `False`. Currently, there's no logic distinguishing environment termination due to a valid end condition (`terminated`) from other reasons, e.g., a time limit (`truncated`). Adding such differentiation is beyond the scope of this PR, unless specifically requested.


Baekalfen:
v2.0.0 removes Gym from the repo. See Wiki for example instead

PR #268, opened 2023-10-29
Remove extra comparisons

Baekalfen:


Baekalfen:
Couldn't prove any performance difference

PR #280, opened 2023-11-02
Multitick

thatguy11325:
Adds a new function, `multitick` which will allow users to tick multiple steps at a time and minimize CPython overhead.


Baekalfen:
Thank you! This has been worked into v2.0.0 :)

PR #288, opened 2023-11-19
Use integer subtraction over modulo arithmetic for a slight speed boost

thatguy11325:
I doubt self.clock or self.clock_target ever exceeds 2*FRAME_CYCLES. In testing, this change showed a 3-6% speed increase assuming the screen is being rendered every tick.


Baekalfen:
I tried benchmarking some more, but it didn't seem to work

Issue #249, opened 2023-06-14
Change keybindings


zanprgmr:
I wanted to suggest an improvement for PyBoy, the Game Boy emulator. It would be great if there was an easy way to change the keybindings. Perhaps you could add a function or a "Wrapper" that allows users to customize the keybindings more easily. This would make PyBoy more user-friendly and accessible.

Additionally, I think it would be awesome if the PyBoy community could get involved in improving the keybindings. It would be amazing to see the community implement controller compatibility, so users can enjoy PyBoy with their preferred controllers. By opening up the keybindings code to community contributions, we can all work together to enhance this aspect of the emulator.

I really appreciate the work you've put into PyBoy so far, and I believe these enhancements would make it even better. Thank you for considering this request, and I'm excited to see what the future holds for PyBoy!

Baekalfen:
It's definitely something that would be nice to have. I know others have worked on it before.

Just to point out. There is controller support already, but it's just not documented: https://github.com/Baekalfen/PyBoy/blob/a464e6949a0681eeda34d3c4d476db19b42aced1/pyboy/plugins/window_sdl2.py#L66

bhaskar253:
@Baekalfen I have a query regarding how to approach this issue, one thing I was thinking to add the mapping to some config file & read it like SDL_GameControllerAddMappingsFromFile( 'gamecontrollerdb.cfg' );
is this approach ok?

tk744:
Here's a solution I've been using.

```python
import pyboy
from pyboy.utils import WindowEvent
from pyboy.plugins.window_sdl2 import sdl2

if pyboy.plugins.window_sdl2.sdl2:
    pyboy.plugins.window_sdl2.KEY_DOWN[sdl2.SDLK_j] = WindowEvent.PRESS_BUTTON_A
    pyboy.plugins.window_sdl2.KEY_DOWN[sdl2.SDLK_k] = WindowEvent.PRESS_BUTTON_B
    pyboy.plugins.window_sdl2.KEY_DOWN[sdl2.SDLK_w] = WindowEvent.PRESS_ARROW_UP
    pyboy.plugins.window_sdl2.KEY_DOWN[sdl2.SDLK_s] = WindowEvent.PRESS_ARROW_DOWN
    pyboy.plugins.window_sdl2.KEY_DOWN[sdl2.SDLK_a] = WindowEvent.PRESS_ARROW_LEFT
[...]

PR #289, opened 2023-11-27
Change keybindings for sdl2

tk744:
Added feature to change the sdl2 keybindings as suggested by [this issue](https://github.com/Baekalfen/PyBoy/issues/249).

Usage:
```python
from pyboy.plugins.window_sdl2 import set_keybinding

set_keybinding(sdl2.SDLK_w, WindowEvent.PRESS_ARROW_UP, WindowEvent.RELEASE_ARROW_UP)
set_keybinding(sdl2.SDLK_s, WindowEvent.PRESS_ARROW_DOWN, WindowEvent.RELEASE_ARROW_DOWN)
set_keybinding(sdl2.SDLK_a, WindowEvent.PRESS_ARROW_LEFT, WindowEvent.RELEASE_ARROW_LEFT)
set_keybinding(sdl2.SDLK_d, WindowEvent.PRESS_ARROW_RIGHT, WindowEvent.RELEASE_ARROW_RIGHT)
```

### DIFF
--- a/pyboy/plugins/window_sdl2.py
+++ b/pyboy/plugins/window_sdl2.py
@@ -95,6 +95,10 @@ else:
     CONTROLLER_UP = {}
 # yapf: enable
 
+def set_keybinding(sdl2_key, press_window_event, release_window_event):
+    if sdl2:
+        KEY_DOWN[sdl2_key] = press_window_event
+        KEY_UP[sdl2_key] = release_window_event
 
 def sdl2_event_pump(events):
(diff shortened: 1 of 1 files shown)


Baekalfen:
Thank you for your PR! I'll have to consider it. I would have preferred if it was a generic solution for OpenGL and a future window type too, but I don't have a good idea on how to do it. Problem is that once it's in, I'll have to keep supporting it going forward.

Issue #260, opened 2023-10-24
Only show frame-limiting warning when calling set_emulation_speed


Baekalfen:
Headless, dummy and OpenGL windows will report the following when initialized. This should instead be triggered when calling set_emulation_speed.

```
pyboy.plugins.window_headless  WARNING  This window type does not support frame-limiting.
```

PR #294, opened 2023-12-05
Only show frame-limiting warning when calling set_emulation_speed

owen-crook:
tackling https://github.com/Baekalfen/PyBoy/issues/260

remove warnings from the individual plugins, add warning into ``set_emulation_speed`` after first initialization if not supported with current window_type


Baekalfen (approved):
Looks good! Thank you :)

PR #295, opened 2023-12-06
Update Sprite palette number to support CGB

NicoleFaye:
I noticed that the palette in the sprite object only supported DMG mode as bit 4 is unused in CGB mode and instead uses bits 2-0 for the palette options

https://gbdev.io/pandocs/OAM

### DIFF
--- a/pyboy/api/sprite.py
+++ b/pyboy/api/sprite.py
@@ -117,7 +117,7 @@ class Sprite:
         """
 
         if self.mb.cgb:
-            self.attr_palette_number = attr & 0x3
+            self.attr_palette_number = attr & 0b111
         else:
             self.attr_palette_number = _bit(attr, 4)
         """
--- a/pyboy/api/sprite.py
+++ b/pyboy/api/sprite.py
@@ -116,12 +116,7 @@ class Sprite:
             The state of the bit in the attributes lookup.
         """
 
-        if self.mb.cgb:
-            self.attr_palette_number = attr & 0b111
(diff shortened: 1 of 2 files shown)


pyboy/api/sprite.py
```diff
@@ -116,15 +116,18 @@ def __init__(self, mb, sprite_index):
             The state of the bit in the attributes lookup.
         """
 
-        self.attr_palette_number = _bit(attr, 4)
+        if self.mb.cgb:
+            self.attr_palette_number = attr & 0x3
```
Baekalfen, line 120: Thank you, I had completely missed that. Generally the API has not been fully updated for CGB.

I think it should actually be `attr & 0x7`, as it's 3 bits https://gbdev.io/pandocs/OAM#byte-3--attributesflags

And I see that the bank attribute is missing too

NicoleFaye: Yeah i was tired when i found it, i was thinking 3 bits and i wrote 3 lol.

NicoleFaye: Take a look at the bank number var i added and see if that implementation makes sense to you.

Baekalfen: Perfect! The system I use for docs is a little weird, so I added a commit to fix that. The docstring has to come right after the variable, and it cannot be in an `if`-statement.

I also almost fell for the 3 vs. 7 haha

Baekalfen (approved):

PR #298, opened 2023-12-17
Memory Scanner squashed

Baekalfen:
Should contain the same as #293 but all commits have been squashed to make review and rebase easier.


pyboy/api/memoryscanner.py
```diff
@@ -0,0 +1,119 @@
...
+        :param target_value: The value to search for or None for any value (used in relation to `MemoryScanner.rescan_memory`).
+        :param standard_comparison_type: The type of comparison to use.
+        :param value_type: The type of value (INT or BCD) to consider.
+        :param byte_width: The number of bytes to consider for each value.
+        :param byteorder: The endian type to use (see [int.from_bytes](https://docs.python.org/3/library/stdtypes.html#int.from_bytes)). Note, this is only used for 16-bit values and higher.
+        :return: A list of addresses where the target value is found.
+        """
+        self._memory_cache = {}
+        self._memory_cache_byte_width = byte_width
+        for addr in range(start_addr, end_addr - byte_width + 2): # Adjust the loop to prevent reading past end_addr
```
Baekalfen, line 66: I'm not sure I see how `range(start_addr, end_addr - byte_width + 2)` works. Shouldn't it be: ` range(start_addr, end_addr+1, byte_width)`?

And I think we should raise an exception if `(end_addr+1)-start_addr` doesn't divide evenly with `byte_width`

pyboy/api/memoryscanner.py
```diff
@@ -0,0 +1,119 @@
...
+        Scans memory in the specified range, looking for a target value.
+
+        :param start_addr: The starting address for the scan.
+        :param end_addr: The ending address for the scan.
+        :param target_value: The value to search for or None for any value (used in relation to `MemoryScanner.rescan_memory`).
+        :param standard_comparison_type: The type of comparison to use.
+        :param value_type: The type of value (INT or BCD) to consider.
+        :param byte_width: The number of bytes to consider for each value.
+        :param byteorder: The endian type to use (see [int.from_bytes](https://docs.python.org/3/library/stdtypes.html#int.from_bytes)). Note, this is only used for 16-bit values and higher.
+        :return: A list of addresses where the target value is found.
```
Baekalfen, line 62: The docs generator we're using doesn't pick up this format. You can see `pyboy.py` for examples on how to format it.

pyboy/api/memoryscanner.py
Baekalfen: Ah, is it because you're scanning all possible alignments, not just say the 2-byte aligned values?

pyboy/api/memoryscanner.py
```diff
@@ -0,0 +1,119 @@
...
+        return list(self._memory_cache.keys())
+
+    def _check_value(self, value, target_value, standard_comparison_type):
+        """
+        Compares a value with the target value based on the specified compare type.
+
+        :param value: The value to compare.
+        :param target_value: The target value to compare against.
+        :param standard_comparison_type: The type of comparison to use.
+        :return: True if the comparison condition is met, False otherwise.
```
Baekalfen, line 107: The docs generator we're using doesn't pick up this format. You can see `pyboy.py` for examples on how to format it.

pyboy/api/memoryscanner.py
```diff
@@ -0,0 +1,119 @@
...
+            if value_type == ScanMode.BCD:
+                value = bcd_to_dec(value, byte_width, byteorder)
+
+            if target_value is None or self._check_value(value, target_value, standard_comparison_type.value):
+                self._memory_cache[addr] = value
+
+        return list(self._memory_cache.keys())
+
+    def rescan_memory(self, dynamic_comparison_type=DynamicComparisonType.UNCHANGED, new_value=None):
+        for addr, value in self._memory_cache.copy().items():
```
Baekalfen, line 80: We should add some docs here too.

pyboy/utils.py
```diff
@@ -276,3 +277,48 @@ def __init__(
...
+
+def dec_to_bcd(value, byte_width=1, byteorder="little"):
+    """
+    Converts a decimal value to Binary Coded Decimal (BCD).
+
+    :param value: Integer value to convert.
+    :param byte_width: The number of bytes to consider for each value.
+    :param byteorder: The endian type to use (see [int.from_bytes](https://docs.python.org/3/library/stdtypes.html#int.from_bytes)). Note, this is only used for 16-bit values and higher.
+    :return: BCD equivalent of the decimal value.
+    """
```
Baekalfen, line 295: The docs generator we're using doesn't pick up this format. You can see `pyboy.py` for examples on how to format it.

pyboy/utils.py
```diff
@@ -276,3 +277,48 @@ def __init__(
...
+
+def bcd_to_dec(value, byte_width=1, byteorder="little"):
+    """
+    Converts a Binary Coded Decimal (BCD) value to decimal.
+
+    :param value: BCD value to convert.
+    :param byte_width: The number of bytes to consider for each value.
+    :param byteorder: The endian type to use (see [int.to_bytes](https://docs.python.org/3/library/stdtypes.html#int.to_bytes)). Note, this is only used for 16-bit values and higher.
+    :return: Decimal equivalent of the BCD value.
+    """
```
Baekalfen, line 314: The docs generator we're using doesn't pick up this format. You can see `pyboy.py` for examples on how to format it.

tests/test_bcd_conversion.py
```diff
@@ -0,0 +1,100 @@
...
+    pyboy.tick(175, True)
+    assert pyboy.memory[score_100] == 4
+    addresses = pyboy.memory_scanner.scan_memory(4, start_addr=0xC000, end_addr=0xDFFF)
+    assert score_100 in addresses
+    values = [pyboy.memory[x] for x in addresses]
+    assert all([x == 4 for x in values])
+
+    pyboy.tick(175, True)
+    assert pyboy.memory[score_100] == 8
+    addresses = pyboy.memory_scanner.rescan_memory(8, DynamicComparisonType.MATCH)
```
Baekalfen, line 68: There seems to be something wrong with `rescan_memory(..., DynamicComparisonType.MATCH)`. From line 62, 64 and 67 we know for certain that the address is in the scanned range, we know that it is equal to 4 and we know it changes to 8. But the rescan returns an empty list.

pyboy/api/memoryscanner.py
NicoleFaye: Yeah, my thought is that there is a chance of odd byte allocation sizes which would make not all values 2 byte aligned i think. I just wanted to make sure no value was overlooked by mistake. Does that logic make sense to you as well?

NicoleFaye: I'm not positive its the best implementation. but i changed it because it crashed when i was using it when working on a pokemon pinball wrapper.

tests/test_bcd_conversion.py
```diff
@@ -0,0 +1,100 @@
...
+    assert len(addresses) == 0
+
+
+score_100 = 0xD072
+
+
+def test_memoryscanner_absolute(kirby_rom):
+    pyboy = PyBoy(kirby_rom, window_type="null", game_wrapper=True)
+    pyboy.set_emulation_speed(0)
+    kirby = pyboy.game_wrapper
```
NicoleFaye, line 52: This is missing ()

tests/test_bcd_conversion.py
NicoleFaye: #299

PR #299, opened 2023-12-17
Fix comment format for squashed memory scanner

NicoleFaye:
Fixed Comments, still need to troubleshoot the MATCH case.


NicoleFaye:
I think this fixes it at least. the returns might not be right.

NicoleFaye:
Okay i added a fix for the rescan method as well, the comparison was wrong and the rescan should update the cache value if it meets the requirement. So it does that now.

PR #305, opened 2024-01-27
Mask Rom Bank To Required Bits

nishantc1527:
image
Taken from [here](https://gbdev.io/pandocs/MBC1.html)

Added this missing functionality to the mbc1 mapper. I am not sure of any games that would require this / break without this. Tests still pass.

### DIFF
--- a/pyboy/core/cartridge/mbc1.py
+++ b/pyboy/core/cartridge/mbc1.py
@@ -26,6 +26,8 @@ class MBC1(BaseMBC):
             # Attempting to write 0b00000 will write 0b00001 instead.
             if value == 0:
                 value = 1
+            romsize = self.rombanks[0, 0x0148]
+            value = value & ((1 << (romsize + 1)) - 1);
             self.bank_select_register1 = value
         elif 0x4000 <= address < 0x6000:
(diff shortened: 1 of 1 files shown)


Baekalfen:
Hi, thank you for your PR! Have you identified any issue with not having this? Because I think it might be covered by the get method and `% self.external_rom_count`.

But I can see that we never check 0x148, but that might be good to do in general.

nishantc1527:
I did not notice the ```external_rom_count``` part, sorry. However, I noticed that this value is gotten from the length of the file. Could this be something else to fix? Because there are cases when the rom file is padded at the end for whatever reason leading to inaccuracies, and reading directly from ```0x0148``` will prevent inaccuracies.

Baekalfen:
> I did not notice the `external_rom_count` part, sorry. However, I noticed that this value is gotten from the length of the file. Could this be something else to fix? Because there are cases when the rom file is padded at the end for whatever reason leading to inaccuracies, and reading directly from `0x0148` will prevent inaccuracies.

No worries :) Would you like to do a quick survey, and check the games you have? It's allowed to be padded with zeros, but I have no idea how often the files don't match the bank count in 0x148.

PR #310, opened 2024-02-23
Move hook initialization

NicoleFaye:
This adjusts the initialization to be before the plugin manager as well. I made a mistake in my last PR and didn't move it to the right line, my apologies.


Baekalfen:
I'll fix this up for you, but I think you should look at Git rebase. It's a good way to avoid merge commits, and it's a really effective tool if you get comfortable using it. Definitely worth the effort :)

PR #311, opened 2024-02-26
Fix wrong addressing in the debugging memory view

LittleNyima:
I noticed that there is something wrong with the addressing of the debugging memory view. In this view, there are 16 values in each row, thus the address should increment by 0x10 each time. However, in the code, the address is incremented by 0x20, which leads to wrong results. This also causes the highest 0x200 bytes of the memory is not able to be displayed. Please refer to the attached image and file changes for more details.

illustration of the fix


Baekalfen (approved):
You're right. Nice find! :)

PR #317, opened 2024-03-19
Refactor breakpoint checking

thatguy11325:
Attempted to refactor the breakpoint/hooks logic to not loop over the breakpoints list and use a table lookup for every call. Hopefully it's faster.


Baekalfen:
Could you split up the commits? I would like to test out the breakpoint changes alone, and cherry-pick them onto master. Then we can look at the Makefile afterwards.

pyboy/core/mb.py
```diff
@@ -122,62 +122,64 @@ def breakpoint_add(self, bank, addr):
...
+        pc = self.cpu.PC
+        bank = None
+        if pc < 0x4000 and not self.bootrom_enabled:
+            bank = 0
+        elif 0x4000 <= pc < 0x8000:
+            bank = self.cartridge.rombank_selected
+        elif 0xA000 <= pc < 0xC000:
+            bank = self.cartridge.rambank_selected
+        elif 0xC000 <= pc <= 0xFFFF or (pc <0x100 and self.bootrom_enabled):
+            bank = -1
```
Baekalfen, line 171: There's a potential bug here. But it's my mistake. In other parts of the code, I've used bank 0 for `0xC000 <= pc <= 0xFFFF`.

I'll take a look at this.

pyboy/core/mb.py
```diff
@@ -122,62 +122,64 @@ def breakpoint_add(self, bank, addr):
...
-        elif addr < 0x4000:
-            self.cartridge.rombanks[bank, addr] = opcode
-        elif 0x4000 <= addr < 0x8000:
-            self.cartridge.rombanks[bank, addr - 0x4000] = opcode
-        elif 0x8000 <= addr < 0xA000:
-            if bank == 0:
-                self.lcd.VRAM0[addr - 0x8000] = opcode
+        if (bank, addr) in self.breakpoints:
+            return [bank, addr, self.breakpoints[(bank, addr)]]
+        return []
```
Baekalfen, line 131: I think tuple is more appropriate here, as this is not mutable data.

pyboy/core/mb.py
```diff
@@ -122,62 +122,64 @@ def breakpoint_add(self, bank, addr):
...
-            self.cartridge.rambanks[bank, addr - 0xA000] = opcode
-        elif 0xC000 <= addr <= 0xE000:
-            self.ram.internal_ram0[addr - 0xC000] = opcode
-        else:
-            logger.error("Unsupported breakpoint address. If this a mistake, reach out to the developers")
-        #     return (None, None, None)
-        # return (bank, addr, opcode)
+                logger.error("Unsupported breakpoint address. If this a mistake, reach out to the developers")
+            #     return (None, None, None)
+            # return (bank, addr, opcode)
```
Baekalfen, line 159: Let's remove these comments, as they're outdated.

pyboy/core/mb.py
```diff
@@ -122,62 +122,64 @@ def breakpoint_add(self, bank, addr):
...
+            opcode = self.breakpoints[(bank, pc)]
+            # Breakpoint hit
+            addr = pc
+            logger.debug("Breakpoint reached: %02x:%04x %02x", bank, addr, opcode)
+            self.breakpoint_waiting = (bank & 0xFF) << 24 | (addr & 0xFFFF) << 8 | (opcode & 0xFF)
+            logger.debug("Breakpoint waiting: %08x", self.breakpoint_waiting)
+            return [bank, addr, opcode]
         logger.debug("Invalid breakpoint reached: %04x", self.cpu.PC)
-        return -1
+        return []
```
Baekalfen, line 181: I think tuple is more appropriate here, as this is not mutable data.

Makefile
```diff
@@ -10,11 +10,12 @@ else
     CFLAGS='-w -DCYTHON_WITHOUT_ASSERTIONS'
 endif
 
-PY := python3
+PY ?= python3
 PYPY := pypy3
```
Baekalfen, line 14: Good idea, maybe add PyPy too?

Makefile
```diff
@@ -70,7 +71,7 @@ test_cpython_doctest:
 	${PY} -m pytest pyboy/ -n auto -v
 
 test_cython:
-	${PY} -m pytest tests/ -n auto -v
+	${PY} -m pytest tests/ ${TEST_CYTHON_ARGS}
 
 test_pypy:
 	${PYPY} -m pytest tests/ pyboy/ -n auto -v
```
Baekalfen, line 77: We could call them `${PYTEST_ARGS}` and add them to PyPy too. They would normally run the same setup.

Baekalfen (changes requested):
Really nice work! And sorry for the late review.

I've commented on everything that came to mind. You're welcome to fix the things you care for, and leave the rest to me.

thatguy11325:
Took care of all the requested changes. I didn't know if you wanted new commits or amended old commits. I went with the latter.

One possible change I saw, but didn't make is changing the breakpoints dict to a fixed-length array of size `0xFFFF` (similar to your trick in your jit experiment). I believe that'd allow removing the gil from `breakpoint_reached`. Using libcpp's `unordered_map` would allow for a similar no-gil, but would create a stricter dependency on cython.

Baekalfen (approved):
Great work! Thank you for doing this! :)

PR #325, opened 2024-05-17
fix pokemon tilemap wapper for 2.0

stangerm2:
<!--

Fix Pokemon wrapper Screen Tile helper method for 2.0:

    2.0 introduced a new method for getting the screen tile map, this
    commit simply swaps the old 1.x API for the new 2.x API per API
    doc: https://docs.pyboy.dk/api/screen.html#pyboy.api.screen.
    Screen.get_tilemap_position

FYI, the screen methods is a bit confusing to me, specifically this part in the screen doc:

>     If you're making an AI or bot, it's highly recommended to _not_ use this class for detecting objects on the screen.
>     It's much more efficient to use `pyboy.PyBoy.tilemap_background`, `pyboy.PyBoy.tilemap_window`, and `pyboy.PyBoy.get_sprite` instead.
> 

It seems like `self.pyboy.screen.get_tilemap_position()` is still required though? Is this still the most efficient way to get the tilemap then? Otherwise, hope it helps.


Baekalfen (approved):
Looks good! Thank you

PR #336, opened 2024-07-21
Expand cartridge information

fxck-tony:
Expand PyBoy API to return more data from the ROM cartridge's header as it region and cartridge type

### DIFF
--- a/pyboy/core/cartridge/base_mbc.pxd
+++ b/pyboy/core/cartridge/base_mbc.pxd
@@ -15,6 +15,8 @@ cdef Logger logger
 cdef class BaseMBC:
     cdef str filename
     cdef str gamename
+    cdef srt gametype
+    cdef str gameregion
     cdef uint8_t[:, :] rombanks
     cdef uint8_t[:,:] rambanks
     cdef uint8_t carttype
@@ -37,6 +39,8 @@ cdef class BaseMBC:
     cdef void load_ram(self, IntIOInterface) noexcept
     cdef void init_rambanks(self, uint8_t) noexcept
     cdef str getgamename(self, uint8_t[:,:]) noexcept
+    cdef str getgametype(self, uint8_t[:,:]) noexcept
+    cdef str getgameregion(self, uint8_t[:,:]) noexcept
 
     cdef uint8_t getitem(self, uint16_t) noexcept nogil
     cdef void setitem(self, uint16_t, uint8_t) noexcept nogil
--- a/pyboy/core/cartridge/base_mbc.py
+++ b/pyboy/core/cartridge/base_mbc.py
@@ -34,6 +34,8 @@ class BaseMBC:
         self.external_ram_count = external_ram_count
         self.init_rambanks(external_ram_count)
         self.gamename = self.getgamename(rombanks)
+        self.gametype = self.getgametype(rombanks)
+        self.gameregion = self.getgameregion(rombanks)
 
         self.memorymodel = 0
         self.rambank_enabled = False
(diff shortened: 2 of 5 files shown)


pyboy/core/cartridge/base_mbc.pxd
```diff
@@ -15,6 +15,8 @@ cdef Logger logger
 cdef class BaseMBC:
     cdef str filename
     cdef str gamename
+    cdef srt gametype
```
Baekalfen, line 18: Misspelled `str`

pyboy/core/cartridge/base_mbc.py
```diff
@@ -103,7 +105,55 @@ def init_rambanks(self, n):
...
-        return "".join([chr(rombanks[0, x]) for x in range(0x0134, 0x0142)]).split("\0")[0]
+        return "".join([chr(rombanks[0, x]) for x in range(0x0134, 0x0143)]).split("\0")[0]
+
+    def getgametype(self, rombanks):
+        if rombanks[0, 0x143] == 0x80 or rombanks[0, 0x143] == 0xc0:
+            return "Game Boy Color"
+        elif rombanks[0, 0x146] == 0x03:
+            return "Super Game Boy"
+        else:
+            return "Game Boy"
```
Baekalfen, line 116: It's probably better to split this out into separate "cgb" and "sgb" checks, as they are seemingly not mutually exclusive https://gbdev.io/pandocs/The_Cartridge_Header.html

pyboy/core/cartridge/base_mbc.py
```diff
@@ -103,7 +105,55 @@ def init_rambanks(self, n):
...
+            return "Unknown"
+        else:
+            # Get the last letter of the sales code that determines the region
+            region_string = "".join([chr(rombanks[0, x]) for x in range(0x13f, 0x0143)]).split("\0")[0]
+            if region_string:
+                region_code = region_string[-1]
+            else:
+                return "Unknown"
+            # Return game region according to the last letter of the serial code
+            if region_code == 'J':
```
Baekalfen, line 131: Where did you find these codes?

pyboy/core/cartridge/base_mbc.py
```diff
@@ -103,7 +105,55 @@ def init_rambanks(self, n):
...
+        elif rombanks[0, 0x146] == 0x03:
+            return "Super Game Boy"
+        else:
+            return "Game Boy"
+
+    def getgameregion(self, rombanks):
+        game_type = self.getgametype(rombanks)
+        # The region character is only present in GBC/SGB games only; for regular GB games can't be fetched
+        if game_type == 'Game Boy':
+            return "Unknown"
```
Baekalfen, line 122: It would be better to return `None`, when we don't have an answer. It's more intuitive that it returns something that evaluates to False: `bool(pyboy.cartridge_region) == False`.

I.e. this won't work if it's `"Unknown"`:
```python
if python.cartridge_region:
    do_something(python.cartridge_region)
```

pyboy/pyboy.py
```diff
@@ -323,6 +323,42 @@ def __init__(
...
+        'Europe'
+
+        ```
+
+        Returns
+        -------
+        str :
+            Europe
+        """
+
```
Baekalfen, line 361: I think it might make sense to move this to `pyboy.cartridge.region` and create a cartridge API, just like `pyboy.screen` etc. I can do it, if you think it's too much.

pyboy/core/cartridge/base_mbc.py
```diff
@@ -103,7 +105,55 @@ def init_rambanks(self, n):
         self.rambanks = memoryview(array.array("B", [0] * (8*1024*16))).cast("B", shape=(16, 8 * 1024))
 
     def getgamename(self, rombanks):
-        return "".join([chr(rombanks[0, x]) for x in range(0x0134, 0x0142)]).split("\0")[0]
+        return "".join([chr(rombanks[0, x]) for x in range(0x0134, 0x0143)]).split("\0")[0]
```
Baekalfen, line 108: Would it make sense to conditionally extract a shorter or longer string if it's a CGB or DMG ROM? I know it's a little outside of your PR

Baekalfen:
Closed for inactivity

Issue #286, opened 2023-11-14
Create Gameshark plugin [Recommended first issue] 


Baekalfen:
*This is a good first PR for someone to get started*

~~This patch should be implemented as a plugin, and maybe fix up what ever it might need to work as a plugin.~~

We don't have the plugin infrastructure in place yet, so make the implementation in the api directory and expose the functions on the PyBoy object (just like screen, sprite, tilemap, etc.). 

https://github.com/Fort-Bonnitar/PyBoy/commit/a8a74b726f69c26c13030f0baf1353f45adff9d8

Baekalfen:
This has already been solved.

PR #341, opened 2024-09-07
add gameShark api

daivikbhatia:
#286 Added support for GameShark API

- Created two new files in the `API` directory: `gameshark.py` and `gameshark.pdx`.
- Added a `run_cheats()` function that can be called from PyBoy using `pyboy.gameshark.run_cheats(cheat, _is_path)`.
- Introduced an _is_path flag in the run_cheats() function:
   - If _is_path is True, cheat must be a path to a text file containing cheats.
   - If _is_path is False, cheat can be passed directly as a string.
 
Example Usage:
```
from pyboy import PyBoy
pyboy = PyBoy("./red.gb")
while pyboy.tick():
    pyboy.gameshark.run_cheats("010138CD",_is_path=False)
    pass
pyboy.stop()
```

### DIFF
--- a/pyboy/api/__init__.py
+++ b/pyboy/api/__init__.py
@@ -11,6 +11,7 @@ from .screen import Screen
 from .sprite import Sprite
 from .tile import Tile
 from .tilemap import TileMap
+from .gameshark import GameShark
 
 # __pdoc__ = {
 #     "constants": False,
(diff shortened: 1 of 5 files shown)


Baekalfen:
Is the cheats file a commonly used format for GameShark cheats?

Otherwise, I'm thinking to just have the API apply one at a time from a string (`_is_path=False`) and remove this complexity. The end-user can easily implement this on their own.

Baekalfen:
Also, I don't use these cheats myself. Could you add a `tests/test_gameshark.py`?

Baekalfen:
Superseded by #345. Thank you so much for the initial work! It's still the basis of the other PR.

Issue #333, opened 2024-06-27
Add frame limiter to `null` window


Baekalfen:
Add frame limiter in the style of the OpenGL implementation https://github.com/Baekalfen/PyBoy/commit/2d8a606fc3596c23111548c916aef016d18a7c82

daivikbhatia:
@Baekalfen I want to take it up

Baekalfen:
Solved in #333

PR #342, opened 2024-09-07
Fixed issue #333: Moved frame limiter to base plugin

daivikbhatia:
#333  Moved frame limiter to base plugin

- Moved the frame limiter function from OpenGL window and SDL2 Window to PyBoyWindowPlugin so it can also run on Null window.
- The frame limiter function should work same as before on OpenGL window and SDL2 Window as both of them can inherit the function from PyBoyWindowPlugin.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1041,7 +1041,8 @@ True...
 
         A `target_speed` of `0` means unlimited. I.e. fastest possible execution.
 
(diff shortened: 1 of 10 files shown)


Baekalfen:
Great job! You have introduced a few `print` statements. If you remove these, I think we're ready to merge

Baekalfen (approved):

PR #337, opened 2024-08-21
Serial Support

Artucuno:
- Merged changes from master branch
- Working implementation of Serial

**Tested in:**
- Pokemon Yellow
  - Trading
  - Battle

**Issues:**
- ~~There is a small issue where the emulator will temporarily freeze when waiting for the other Gameboy to respond to data (Should be an easy fix for someone)~~ - Fixed in 7878718
- Synchronization issues in some games, however Pokémon games work completely fine.
  - Tetris
- When compiled into Cython, PyBoy freezes after socket connection is established

PR #344, opened 2024-09-08
Serial Support

Artucuno:
(Reopened #337)

- Merged changes from master branch
- Working implementation of Serial

**Tested in:**
- Pokemon Yellow **(WORKING)**
  - Trading
  - Battle

**Issues:**
- ~~There is a small issue where the emulator will temporarily freeze when waiting for the other Gameboy to respond to data (Should be an easy fix for someone)~~ - Fixed in 7878718 & fa3eaaa
- ~~Synchronization issues in some games, however Pokémon games work completely fine.~~ Fixed in fa3eaaa
- ~~When compiled into Cython, PyBoy freezes after socket connection is established~~ Fixed in fa3eaaa
- GBs will only use the internal clock


Baekalfen:
Very interesting work! Could you rebase this instead? I can't tell exactly which changes are yours and which are from the merge commit.

Also, you seem to change some of the formatting in the files. You should use `pip install pre-commit && pre-commit install` and let that take care of the formatting.

beardyFace:
@Artucuno how do I test/use the serial connection through pyboy with these changes?

Baekalfen:
Superseded by #430

PR #347, opened 2024-09-17
Small fix to README examples

Artucuno:
Updated no rendering example from
```python
pyboy.tick(target) # render=False not set

```
to
```python
for _ in range(target):
    pyboy.tick(render=False)

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,8 @@ for _ in range(target//15):
 <td >
 
 ```python
-pyboy.tick(target)
+for _ in range(target):
+    pyboy.tick(render=False)
 
 ```
 </td>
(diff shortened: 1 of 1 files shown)


Baekalfen:
Hmm. I mean it does render the last frame, but it will have minimal impact on performance. Tried to strike a balance between simplicity and function for the example.

Maybe the best would be to reduce your example to `pyboy.tick(target, render=False)`?

Artucuno:
> Maybe the best would be to reduce your example to `pyboy.tick(target, False)`?

Good idea. I made the update 👍

Baekalfen (approved):

PR #348, opened 2024-09-17
Add button mapping to README

msosav:


Baekalfen:
Thank you, it's valid feedback. But I try to keep the readme at a minimum, as people tend not to read it, if it gets too long.

These instructions are also printed out when you start the emulator, so I don't find it necessary to put in the readme atm.

PR #349, opened 2024-09-18
BombermanGB Wrapper

NIvo172:
This PR adds a BombermanGB wrapper, that I used to train an agent to play arena mode. 

After calling the `start_game()` method, the game automatically navigates to the arena mode, this includes entering the password to unlock the game mode.

I am open to suggestions, if you see anything I could improve.


Baekalfen:
Very cool! Could you add some tests too? Just to check the basics to make sure I don't accidentally break compatibility later.

PR #360, opened 2025-01-31
GameWrapper: Add second Tile for Sprites using two Tiles in the game area matrix

BackrndSource:
This update ensures that `Sprites` consisting of two `Tiles` are fully added in the game area matrix provided by the GameWrapper plugin in `pyboy.game_area()` or `pyboy.game_wrapper.game_area()`

### Bug fixed:

- Only the first tile was placed, leading to incomplete game area representations depending on the game.

###  Why Is It Important?

When training an AI model, the most common approach is to directly use game_area as a parameter. Some games use sprites composed of two tiles. Since understanding the Game Boy's sprite and tile system is not necessarily required for training a model, this issue can be difficult to debug.

As a result, those who notice something wrong with game_area might not realize that this is the root cause of the problem. This could have a significant impact while going unnoticed.

####  Another question:

I have some minor modifications for the GameWrapper that do not affect its behavior but improve the code in terms of readability and use of constants. It’s only a few lines, and you can check the code in this branch of my fork: [https://github.com/BackrndSource/PyBoy/tree/GameWrapper__pretty](https://github.com/BackrndSource/PyBoy/tree/GameWrapper__pretty)

Should I add it to this pull request or create a new one?

### DIFF
--- a/pyboy/plugins/base_plugin.py
+++ b/pyboy/plugins/base_plugin.py
@@ -261,9 +261,12 @@ class PyBoyGameWrapper(PyBoyPlugin):
         for s in sprites:
             _x = (s.x // 8) - xx
             _y = (s.y // 8) - yy
-            if 0 <= _y < height and 0 <= _x < width:
-                tiles_matrix[_y][_x] = self.mapping[
-                    s.tile_identifier] + self.sprite_offset # Adding offset to try to seperate sprites from tiles
+            if 0 <= _x < width:
+                if 0 <= _y < height:
+                    tiles_matrix[_y][_x] = self.mapping[s.tile_identifier] + self.sprite_offset
+                if len(s.tiles) == 2 and 0 <= _y + 1 < height:  # Sprite has two tiles
+                    tiles_matrix[_y + 1][_x] = self.mapping[s.tile_identifier + 1] + self.sprite_offset
+
         return tiles_matrix
 
     def game_area_mapping(self, mapping, sprite_offest):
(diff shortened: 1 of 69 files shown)


Baekalfen:
Thank you for your PRs! I'll take a look as soon as possible. Last two weeks has been quite busy for me

pyboy/plugins/base_plugin.py
```diff
@@ -264,6 +264,9 @@ def game_area(self):
             if 0 <= _y < height and 0 <= _x < width:
                 tiles_matrix[_y][_x] = self.mapping[
                     s.tile_identifier] + self.sprite_offset # Adding offset to try to seperate sprites from tiles
+                if _y + 1 < height and len(s.tiles) == 2:
+                    tiles_matrix[_y + 1][_x] = self.mapping[s.tile_identifier + 1] # Adding second tile
```
Baekalfen, line 268: Nice catch! I think we need to handle the corner case where a sprite is placed above the screen, and only the lower tile is visible. That would have been filtered out by the `0 <= _y` above.

In `lcd.py` we check that the scanline is between `y` and `y+16`.

Baekalfen (changes requested):

Baekalfen:
Saw your comment about the other patch. Just push both commits to this PR

pyboy/plugins/base_plugin.py
BackrndSource: In the latest pull, it's solved at least for the case where the indicated game area cuts the sprite in half. However, I'm not sure if a sprite can be placed at position -1 if the bounds are defined by the screen itself.

Baekalfen: Perfect! I've added a test case which demonstrates that it works too (you're still author of the commits and get credit).

I had to pull in some changes from my WIP branch, as the formatter I used is no longer supported.

Baekalfen (approved):

PR #361, opened 2025-01-31
GameWrapper: Castlevania - The Adventure (GB) 

BackrndSource:
This PR adds a Game Wrapper for Castlevania: The Adventure (GB).


Baekalfen (commented):
Looks good. I'll see if it compiles etc. And then I'll probably have to make a faux test, as I don't own this game. Maybe I'll find it in time.

BackrndSource:
Thank you very much for the code review. I will probably add some methods like set_health or set_lives_left, and maybe other attributes. I can also take the opportunity to create the "vrai" tests. So if you want, don't worry about merging until further notice :)

PR #363, opened 2025-02-10
Minor test fixing

BackrndSource:
Fixing tests:

- `test_log_level_default`: test name is repeated
- `test_game_wrapper_sprites_tall`:  `PyBoyGameWrapper.game_area()` should work without rendering screen


Baekalfen:
Good catch, I actually have the test fixed on another branch that has gotten way too big. I'll try to pick out the common commits and merge them separately from the bigger changes. Bear with me, I'll get them out ASAP.

Baekalfen (approved):

PR #366, opened 2025-02-11
Update WindowEvent.__eq__

BackrndSource:
This ensures proper delegation to `other.__eq__` in case of comparisons with unrelated types and avoid raise `AttributeError` when trying to compare with objects not having `event` attribute

### DIFF
--- a/pyboy/utils.py
+++ b/pyboy/utils.py
@@ -351,11 +351,12 @@ class WindowEvent:
     def __init__(self, event):
         self.event = event
 
-    def __eq__(self, x):
-        if isinstance(x, int):
-            return self.event == x
-        else:
-            return self.event == x.event
+    def __eq__(self, other):
+        if isinstance(other, WindowEvent):
+            return self.event == other.event
+        elif isinstance(other, int):
+            return self.event == other
+        return NotImplemented
 
     def __int__(self):
         return self.event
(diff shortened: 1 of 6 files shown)


pyboy/utils.py
```diff
@@ -243,16 +224,17 @@ class WindowEvent:
...
-        if isinstance(x, int):
-            return self.event == x
-        else:
-            return self.event == x.event
+    def __eq__(self, other):
+        if isinstance(other, WindowEvent):
+            return self.__event == other.__event
+        elif isinstance(other, int):
+            return self.__event == other
+        return NotImplemented
```
Baekalfen, line 234: I see it's correct to do `return NotImplemented`. Can you write a test case that demonstrates it?

tests/test_utils.py
```diff
@@ -0,0 +1,40 @@
+import pytest
+
+from pyboy.utils import WindowEvent
+
+
```
Baekalfen, line 5: I removed the initial test case, as it was testing internal functionality. I've corrected this by changing the code where the internal variable was erroneously used.

pyboy/utils.py
Baekalfen: Possibly, add a test case that demonstrates that not only `a == b` works, but also `b == a`. This is probably done at the same time.

Baekalfen (approved):

Issue #378, opened 2025-03-23
Minor wrapper bug, Pokemon Pinball


ghost:
Pokemon Pinball Gen1's masterball is a 5x mult, and is 5. this causes a print(wrapper) to fail with "no ball_type = 5", minor issue but just wanted to mention.

https://github.com/Baekalfen/PyBoy/blame/master/pyboy/plugins/game_wrapper_pokemon_pinball.py#L231

Baekalfen:
Thank you for reporting. Would you be able to make a PR with a fix?

Ping: @NicoleFaye

PR #381, opened 2025-04-01
Fix Masterball Value In Pokemon Pinball Wrapper

NicoleFaye:
Updated Value for masterball to match actual value found [here](https://github.com/pret/pokepinball/blob/master/constants/ball_types.asm)


NicoleFaye:
Fixing #378

Baekalfen (approved):
Fantastic! Thank you!

PR #382, opened 2025-04-07
Fixed pokemon pinball start game method and associated tests

NicoleFaye:
After messing around with the wrapper for a bit i noticed the start game method wasnt working how i expected it to. Noticed there were changes to the tests we had before. I changed the test values back on 2 tests now that the start game method works as expected.

It was not registering one of the button releases, I assume due to the timing changes since I last worked with it.


Baekalfen (approved):
Fantastic! Thank you

PR #384, opened 2025-04-09
Fix/sound api and recorder

Jdeandrade22:


Jdeandrade22:
the 2 issues 
217
368
should both be fixed and tested with this update

Baekalfen:
Hi, thank you for your PR. You'll need to clean it up a bit. You've commited your virtual env etc. Also please go through each file and remove the things that are not explicitly needed -- like pinning the requirements and adding a comment about AI to `setup.py`

PR #386, opened 2025-04-15
Pokemon pinball wrapper update

NicoleFaye:
Added reset_tracking method to reset all the values being tracked via hooks, allows for more easier management of more diverse training, allowing the ability to create episodes out of a ball until loss instead of life loss or game over.


Baekalfen (approved):
Looks great

Issue #356, opened 2024-11-29
Alleyway: dpad doesn't work


Baekalfen:
Could be a difficult ROM. Other emulators have problems with it too: https://github.com/visualboyadvance-m/visualboyadvance-m/issues?q=Alleyway+

Reported on Discord: https://discord.com/channels/584149554132418570/1312089736164409425/1312089736164409425

not-chciken:
Some games, such as Alleyway, require a basic implementation of the serial module to work properly.
The basic requirements include the serial transfer completed interrupt and the correct value of register SB at 0xFF02 (returning 0xFF for a disconnected link cable already suffices ([source](https://gbdev.io/pandocs/Serial_Data_Transfer_(Link_Cable).html#disconnects))).
Pull request #394 fixes this.

Baekalfen:
Great work, thank you!

PR #394, opened 2025-05-11
Adding Serial

not-chciken:
Adding a simple serial module that fixes Issue #356.
It doesn't do much except for scheduling interrupts and returning 0xFF (the default value of a disconnected link cable).
This may seem superfluous but even in single-player mode some games require a serial module to be implemented.

One may also integrate the already existing `serialbuffer` into the `Serial` class but I leave this design choice to the maintainers.


Baekalfen:
Really great work!

I went ahead and did a few fixups. There's something about the `_cycles_to_interrupt` that needs some work. When the serial is not active it should be completely passive, right? As you can see in other places like `timer.py`, I'm assigning `_cycles_to_interrupt = MAX_CYCLES` whenever interrupts are not needed. Can you find a way to add this to Serial?

PR #411, opened 2026-02-17
Add DMG palette cycling with 'C' key

Santitub:
### What does this PR do?

Adds a palette cycling feature for DMG mode. Users can press 'C' 
to switch between 4 predefined color palettes during emulation.

### Available palettes:
- Gray (Default) — Original PyBoy palette
- Classic Green (DMG) — Original Game Boy green tint
- Parchment — Warm beige tones, easy on the eyes
- Mossy — Muted green with earthy tones

### Changes:
- `pyboy/utils.py` — Added CYCLE_PALETTE event
- `pyboy/pyboy.py` — Added palette list, cycling method, event handling
- `pyboy/plugins/window_sdl2.py` — Mapped 'C' key
- `tests/test_palette_cycling.py` — Added tests

### Notes:
- Only works in DMG mode. CGB mode is intentionally ignored
- No changes to lcd.py or any core emulation logic
- Tested with pre-commit hooks


pyboy/pyboy.py
```diff
@@ -252,6 +266,13 @@ def __init__(
             randomize=randomize,
         )
 
+        self._current_palette_idx = 0
+        # Check if the initial palette matches any predefined one
+        for i, pal in enumerate(DMG_PALETTES):
+            if pal == color_palette:
+                self._current_palette_idx = i
+                break
```
Baekalfen: Wouldn't this be an obvious use of a dictionary instead?

pyboy/pyboy.py
```diff
@@ -281,6 +302,12 @@ def __init__(
         # Absolute frame count of the emulation
         self.frame_count = 0
 
+        self._current_palette_idx = 0
+        for i, pal in enumerate(DMG_PALETTES):
+            if pal == color_palette:
+                self._current_palette_idx = i
+                break
```
Baekalfen: Duplicate code

pyboy/pyboy.py
```diff
@@ -617,7 +644,41 @@ def tick(self, count=1, render=True, sound=True):
...
+    
+    def _cycle_palette(self):
+        """Cycles to the next DMG palette."""
+        if self.mb.cgb:
+            logger.info("Palette cycling is only for DMG mode")
+            return
 
+        self._current_palette_idx = (self._current_palette_idx + 1) % len(DMG_PALETTES)
+        new_palette = DMG_PALETTES[self._current_palette_idx]
+        palette_name = DMG_PALETTE_NAMES[self._current_palette_idx]
```
Baekalfen: Dictionaries are ordered in Python. I'd think we could make something with iterators that is more elegant

pyboy/pyboy.py
```diff
@@ -617,7 +644,41 @@ def tick(self, count=1, render=True, sound=True):
...
+            r = (color >> 16) & 0xFF
+            g = (color >> 8) & 0xFF
+            b = color & 0xFF
+            bgr_palette.append((a << 24) | (b << 16) | (g << 8) | r)
+
+        # Update the 3 palette registers: BGP, OBP0, OBP1
+        for palette_reg in [self.mb.lcd.BGP, self.mb.lcd.OBP0, self.mb.lcd.OBP1]:
+            # Replace the 4 base colors of the palette
+            for i in range(4):
+                palette_reg.palette_mem_rgb[i] = bgr_palette[i]
```
Baekalfen: You don't need a range, you can iterate `bgr_palette` directly.

pyboy/pyboy.py
```diff
@@ -617,7 +644,41 @@ def tick(self, count=1, render=True, sound=True):
...
+            # Force lookup table recalculation (bypass the "if value == value" check)
+            old_value = palette_reg.value
+            palette_reg.value = -1
+            palette_reg.set(old_value)
```
Baekalfen: Create a helper-function in `PaletteRegister` that takes the palette and forces the update internally.

tests/test_palette_cycling.py
```diff
@@ -0,0 +1,71 @@
...
+if not os.path.isfile(DEFAULT_ROM):
+    # PyBoy includes a default ROM, find it
+    import pyboy as _pyboy_module
+
+    _pyboy_dir = os.path.dirname(_pyboy_module.__file__)
+    for root, dirs, files in os.walk(_pyboy_dir):
+        for f in files:
+            if f.endswith(".gb"):
+                DEFAULT_ROM = os.path.join(root, f)
+                break
```
Baekalfen: Take a look at how the other tests use `default_rom`

tests/test_palette_cycling.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@pytest.fixture
+def pyboy_dmg():
+    pyboy = PyBoy(DEFAULT_ROM, window="null", sound_emulated=False)
+    pyboy.set_emulation_speed(0)
+    for _ in range(60):
+        pyboy.tick()
+    yield pyboy
+    pyboy.stop(save=False)
```
Baekalfen: It's better to do it "wrong" like all other tests, than to improve just a single test.

Otherwise the PR should be a refactor of all tests.

tests/test_palette_cycling.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def test_cycle_palette_changes_index(pyboy_dmg):
+    assert pyboy_dmg._current_palette_idx == 0
+    pyboy_dmg._cycle_palette()
+    assert pyboy_dmg._current_palette_idx == 1
+    pyboy_dmg._cycle_palette()
+    assert pyboy_dmg._current_palette_idx == 2
```
Baekalfen: As for testing strategy, it's important to test boundary conditions. The middle of a range is not as interesting.

Always test the initial case, and the end case -- i.e. the wrap around. That's where the bugs usually happen.

tests/test_palette_cycling.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def test_cycle_palette_wraps_around(pyboy_dmg):
+    for _ in range(len(DMG_PALETTES)):
+        pyboy_dmg._cycle_palette()
+    assert pyboy_dmg._current_palette_idx == 0
```
Baekalfen: Merge this with the one above

tests/test_palette_cycling.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def test_cycle_palette_updates_bgp(pyboy_dmg):
+    old_colors = [pyboy_dmg.mb.lcd.BGP.palette_mem_rgb[i] for i in range(4)]
+    pyboy_dmg._cycle_palette()
+    new_colors = [pyboy_dmg.mb.lcd.BGP.palette_mem_rgb[i] for i in range(4)]
+    assert old_colors != new_colors
```
Baekalfen: Rather check this end-to-end, through a screen image. You can check a few pixel values. That is what counts. We don't want to test specific internal state of an object.

Also, this wouldn't work when PyBoy is compiled. Screen images will.

tests/test_palette_cycling.py
```diff
@@ -0,0 +1,71 @@
...
+    pyboy_dmg.send_input(WindowEvent.CYCLE_PALETTE)
+    pyboy_dmg.tick()
+    assert pyboy_dmg._current_palette_idx == old_idx + 1
+
+
+def test_cycle_palette_ignored_in_cgb(pyboy_dmg):
+    # Skip if already in CGB mode, or simulate CGB flag
+    if not pyboy_dmg.mb.cgb:
+        # Temporarily set CGB flag to test the guard
+        pyboy_dmg.mb.cgb = True
```
Baekalfen: Don't do this. Use the `any_rom_cgb` fixture.

Baekalfen (changes requested):
Thanks for the PR! It's a good beginning, but it needs a little work

Santitub:
I think I fixed all that you mentioned

Baekalfen (approved):
Sorry I hijacked your PR. I wanted to get some other changes in before your commit, so I didn't have to redo the changes.

Thanks for your contribution!

PR #414, opened 2026-05-02
fix: persistent src/dest tracking during gdma

### LINKED ISSUE #412: Jumbled graphics in Polished Crystal Pokédex
Firstly, **thank you so much** for the work on this amazing project, as it has inspired me to start a little personal side-project using PyBoy as a foundation. :)

This issue might be a long shot as it is based on a ROM hack, but since it's **built from a fully disassembled ROM** I figured I might as well post this as it may be easier to investigate from the ASM + Symbols file and even help unravel issues that should eventually impact other ROMs.

## Context
Pokémon Polished Crystal is a custom game based on the Pokémon Crystal disassembly project. It mostly runs great on PyBoy, however some of the custom features often come with unexpected behavior.

The Pokedex in this game uses a custom implementation for the graphic interface. Here's where most of it is set up: https://github.com/Rangi42/polishedcrystal/blob/254867ed4a663fea553893066b184e8332e713eb/engine/pokedex/lcd.asm#L126

## Behavior (Using pyboy 2.7.0)
https://github.com/user-attachments/assets/e0e08e01-5856-4876-abce-88581042ac7e

Input and logic still work fine, it's just the graphics that appear messed up.

## Expected

Taken from [SameBoy](https://github.com/LIJI32/SameBoy) emulator. Works the same in BGB or mGBA as well.

Image

## How to reproduce
1. Obtain the ROM using patch version 3.2.3 or further.
 
2. Place this RAM file next to your ROM, renaming it so filenames match: [PyBoy_PPC_Pokedex_Graphic_Issue_RAM.zip](https://github.com/user-attachments/files/25458639/PyBoy_PPC_Pokedex_Graphic_Issue_RAM.zip)

3. Boot up the game
```
python -m pyboy -w SDL2 roms/polishedcrystal.gbc
```

4. Hit START and then continue. Once in-game, hit START once more and then A to enter Pokedex. The issue will be visible then.

I will gladly contribute to any further investigation or development, even though my knowledge of the GB's architecture and inner workings are still quite limited, please let me know if there's any way I can help.

Thank you once more.

gVirtu:
## Related issue
closes #412 

## Context
It seems that the problem was directly related to this part:

https://github.com/Baekalfen/PyBoy/blob/8015851f5cbf0a21e837eaf3824e1bad3df7547d/pyboy/core/mb.py#L776-L791

AFAIU, we always base the src/dest off of hdma1-4, which is fine most of the time but might cause problems if a game does multiple sequential transfers without re-writing to these registers.

One approach would be to update these registers to point to the next address instead of 0xFF in lines 788-791, but that seemed to stray from the documented hw behavior so I opted to track them internally with the existing curr_src/curr_dst vars instead.

## Changes made
 - curr_src / curr_dst are now set as soon hdma1-4 are written via setitem_io_ports, in the same way the former src/dst local variables in set_hdma5 used to do
 - in the gdma transfer type:
   - we still set hdma5 to 0xff as per [gbdocs](https://gbdev.io/pandocs/CGB_Registers.html#bit-7--0--general-purpose-dma)
   - we now also persist the next src/dst addresses based on where we left off
 - in the hblank transfer:
   - removed curr_src / curr_dst assign considering they are now already set in setitem_io_ports  

## Tests
Polished Crystal now has the expected behavior described in the issue: the pokédex renders nicely now.

Apparently, no regressions in the test suite, but I also couldn't figure out a good way to unit test this change meaningfully. (will gladly accept any pointers on that :) )

### DIFF
--- a/pyboy/core/mb.py
+++ b/pyboy/core/mb.py
@@ -678,12 +678,16 @@ class Motherboard:
                 self.lcd.vbk.set(value)
             elif self.cgb_mode and i == 0xFF51:
                 self.hdma.hdma1 = value
+                self.hdma.curr_src = (value << 8) | (self.hdma.curr_src & 0x00FF)
             elif self.cgb_mode and i == 0xFF52:
                 self.hdma.hdma2 = value  # & 0xF0
+                self.hdma.curr_src = (self.hdma.curr_src & 0xFF00) | (value & 0xF0)
             elif self.cgb_mode and i == 0xFF53:
                 self.hdma.hdma3 = value  # & 0x1F
+                self.hdma.curr_dst = ((value & 0x1F) << 8) | (self.hdma.curr_dst & 0x00FF) | 0x8000
             elif self.cgb_mode and i == 0xFF54:
                 self.hdma.hdma4 = value  # & 0xF0
+                self.hdma.curr_dst = (self.hdma.curr_dst & 0xFF00) | (value & 0xF0)
             elif self.cgb_mode and i == 0xFF55:
                 self.hdma.set_hdma5(value, self)
                 self.cpu.bail = True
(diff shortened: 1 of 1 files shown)


Baekalfen (approved):
Great work! I’ve always had a suspicion there was something with the DMA

Baekalfen:
To answer your questions about unit testing. You can find inspiration here: https://github.com/Baekalfen/PyBoy/blob/913147b01676369c19150a7ca16c8f81a095d189/tests/test_lcd.py#L23

To test something internal like that, requires we have not compiled with cython. You can instantiate the HDMA object directly too.

Baekalfen:
This example might be better. I instantiate the LCD module without PyBoy https://github.com/Baekalfen/PyBoy/blob/913147b01676369c19150a7ca16c8f81a095d189/tests/test_lcd.py#L154

gVirtu:
Thanks! I noticed one relevant SameSuite test (`gbc_dma_cont`) broke upon merging.

According to [GBEmulatorShootout](https://gbdev.io/GBEmulatorShootout/) we currently weren't rendering the expected output on that one

image

With the change, we're _almost_ there, just gotta figure out why those stray "5"s are showing up down there

image

Baekalfen:
The GBE shootout has sadly always been wrong, and now it’s very, very outdated. I tried correcting the test method when it came out, but the author wasn’t cooperative.

As for why it fails after merging, I don’t know. But it’s probably a quick fix. I can do that next time I’m at a computer.

The “5” you see is the remainder of the PyBoy boot-ROM. If you can load the original boot-ROM, they disappear

Baekalfen:
Oh, I see you posted that we succeed the test now! That’s fantastic! Then I just have to update the screenshot I use to compare — I don’t directly interpret the result, but rather look for change.

PR #418, opened 2026-05-14
Add game wrapper for 2048 (homebrew)

poullakkos:
`Adds a game wrapper for the 2048 homebrew Game Boy game by wyattferguson.`
 **MD5:**  ` 6748719720d57a7dce48d07b2f3c5ede `
 **SHA256:**  ` a5dd7ad505baf13bdf66736518c2d6708459c3aaf1ed2d2282f05c2d1d958d9c `

The wrapper provides:
- score
- board (5x5 grid)
- winner (reached 2048)
- game_over()

Tests included in tests/test_2048.py


Baekalfen:
Awesome, I’ll get back to you asap, but it’ll take some days

Baekalfen:
Can you point me to the ROM you used? The one [here](https://github.com/wyattferguson/2048-gb) has a different MD5: `6748719720d57a7dce48d07b2f3c5ede`

poullakkos:
Thanks for pointing that out! It seems I had an older version of the ROM. I've updated the wrapper to use the official ROM from wyattferguson's repository (MD5: ` 6748719720d57a7dce48d07b2f3c5ede `). The memory addresses are the same so the wrapper works correctly.

pyboy/plugins/manager.py
bhaskar253, line 1: several unrelated formatting changes have been added, do see if these can be cleaned up.

pyboy/plugins/manager_gen.py
```diff
@@ -216,7 +217,8 @@ def skip_lines(iterator, stop):
                 for p in sorted(list(set(plugins) - set(game_wrappers))):
                     p_name = to_snake_case(p)
                     spec = importlib.util.spec_from_file_location(
-                        p_name, os.path.dirname(os.path.abspath(__file__)) + "/" + p_name + ".py"
```
bhaskar253, line 219: here as well unrelated formatting changes have been introduced

tests/conftest.py
bhaskar253, line 1: unrelated formatting changes have been added, please remove them

bhaskar253 (commented):
several unrelated formatting changes have been added, please do avoid them

poullakkos:
Thanks for the feedback! The formatting issues were caused by the Black formatter I had enabled in my editor. I've disabled it and fixed all the formatting. Sorry for the noise!

pyboy/plugins/game_wrapper2048.py
```diff
@@ -0,0 +1,87 @@
...
+def enabled(self):
+    import hashlib
+
+    try:
+        with open(self.pyboy.gamerom_file, "rb") as f:
+            rom_hash = hashlib.md5(f.read()).hexdigest()
+        return rom_hash == "6748719720d57a7dce48d07b2f3c5ede"
+    except Exception as e:
+        logger.error(f"Error occurred while checking ROM hash: {e}")
+        return False
```
Baekalfen, line 26: I can see the game ROM is missing a title, and that's why you do this. I've tried raising the issue with the author: https://github.com/wyattferguson/2048-gb/pull/1

pyboy/plugins/game_wrapper2048.py
```diff
@@ -0,0 +1,87 @@
...
+        self._sprite_cache_invalid = True
+
+        low = self.pyboy.memory[0xC0EF]
+        high = self.pyboy.memory[0xC0F0]
+        self.score = low + (high * 256)
+
+        self.winner = self.pyboy.memory[0xC0F3] == 1
+
+        state = self.pyboy.memory[0xC0F4]
+        self._game_over = state == 3
```
Baekalfen: What is "3"? You should define this magic number as a constant. Are there other interesting states to document?

Baekalfen: It would be nice if you made all of your addresses constants too, with a meaningful name.

Baekalfen (commented):
Short and sweet. Really nice work. Nothing functionally needed (if tests pass).

I've add the 2048 ROM to my mirror and added it to the conftest, so it automatically downloads it.

Baekalfen:
I can see the tests are failing, but you can find the results in the runner outputs

EDIT:
Maybe the output is not public, but you can run it locally

poullakkos:
I initially thought both ROMs were using the same memory addresses, but eventually the first ROM has was using different memory addresses and my wrapper was still using  the old memory addresses. This is why the tests were failing in CI. I have redone the memory mapping with the correct ROM from the PyBoy mirror and updated all memory addresses. Tests now pass locally and should pass CI/CD.

Baekalfen:
Makes sense. Wonder how stable the memory mapping is? At least we are mapping correctly for this specific hash.

Most of the tests pass now, just a single one is failing:
```
    def test_2048_basics(gb2048_file):
        pyboy = PyBoy(gb2048_file, window="null")
        pyboy.set_emulation_speed(0)
    
        wrapper = pyboy.game_wrapper
    
        for _ in range(500):
            pyboy.tick(1, False)
    
        pyboy.button("start")
        for _ in range(60):
            pyboy.tick(1, False)
    
        assert wrapper.score == 0
        assert wrapper.winner == False
        assert wrapper.game_over() == False
    
        non_zero = sum(1 for row in wrapper.board for tile in row if tile != 0)
>       assert non_zero == 2
E       assert 3 == 2

tests/test_2048.py:27: AssertionError
```

pyboy/plugins/game_wrapper2048.py
```diff
@@ -0,0 +1,99 @@
...
+STATE_TITLE = 0
+STATE_PLAYING = 1
+STATE_WINNER = 2
+STATE_GAMEOVER = 3
+
+
+class GameWrapper2048(PyBoyGameWrapper):
+    """
+    This class wraps 2048 for Game Boy, and provides easy access for AIs.
+
```
Baekalfen, line 44: Maybe put something like:

> This game wrapper only works with the release of 2048 with MD5 `6748719720d57a7dce48d07b2f3c5ede`

poullakkos:
Spot on! I had accidentally left the wrong memory address for the start of the board while I was testing different offsets. I've corrected it to the right one now, so hopefully, this last test will finally pass.

I also updated the docstring with the MD5 hash requirement as you suggested. Thanks for the guidance!

Baekalfen (approved):
Great work! I think we’re ready to merge!

PR #420, opened 2026-07-07
Prevent screen recordings from overwriting existing files

yashkuceriya:
## Summary
- reserve screen recording filenames before recording starts
- avoid overwriting existing GIF/MP4 recordings with the same timestamp
- clean up reserved files when recording startup fails

## Tests
- python3 -c 'import sys, types, pytest; sys.modules["git"] = types.ModuleType("git"); sys.exit(pytest.main(["-q", "-o", "addopts=", "tests/test_screen_recorder.py", "tests/test_replay.py"]))'
- python3 -m ruff check pyboy/plugins/screen_recorder.py tests/test_screen_recorder.py tests/test_replay.py
- python3 -m ruff format --check pyboy/plugins/screen_recorder.py tests/test_screen_recorder.py tests/test_replay.py

### DIFF
--- a/pyboy/plugins/screen_recorder.py
+++ b/pyboy/plugins/screen_recorder.py
@@ -3,6 +3,7 @@
 # GitHub: https://github.com/Baekalfen/PyBoy
 #
 
+import contextlib
 import os
 import shutil
 import subprocess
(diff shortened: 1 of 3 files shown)


Baekalfen:

Baekalfen:
I'm not sure I understand the issue. Isn't it currently timestamped? Would anyone start, stop and start a recording within a second?

PR #421, opened 2026-07-15
Add MBC30 support + serial busy-bit fix (#222)

### LINKED ISSUE #222: Japanese Pokemon Crystal: Doesn't start, locks at blank screen
European version works.

fsegouin:
Fixes #222 

## 1. Serial transfers never read as complete

`Serial.tick` masked SC with `0x80` on transfer completion, which *keeps* bit 7 (transfer in progress) set instead of clearing it. Per [Pan Docs](https://gbdev.io/pandocs/Serial_Data_Transfer_(Link_Cable).html), hardware clears bit 7 when the transfer completes, and games poll that bit to wait for completion.

This is the actual cause of the blank screen: Crystal (JP) has all the Mobile Adapter GB code, and its boot-time mobile init performs serial transfers and spins on SC bit 7 forever.

## 2. Missing MBC30 support

Crystal (JP) uses the MBC30, which shares cartridge type codes with MBC3 (`0x0F`–`0x13`) but supports 8 external RAM banks (64KB) and an 8-bit ROM bank register (256 banks / 4MB). The only way to tell it apart is the ROM/RAM size bytes in the header, so `load_cartridge` now selects a new `MBC30` class for MBC3-type carts declaring >4 RAM banks or >128 ROM banks.

Previously these carts ran as plain MBC3, whose write handler caps writable SRAM at bank 3: writes to banks 4–7 were silently dropped while reads worked. The game boots without this fix, but save data written to the upper banks was silently lost.

## Tests

- `tests/test_serial.py` starts an internal-clock transfer through the public memory API and asserts SC bit 7 reads busy immediately, then clears within a frame. Fails on the old code (bit 7 stayed set forever).
- `tests/test_mbc30.py`
- `test_mbc30_detection` synthetic ROM headers in memory, asserting `load_cartridge` selects MBC3 vs. MBC30 across cartridge types `0x10`/`0x13`, RAM size codes `0x03`/`0x05`, and a 256-bank ROM.
- `test_mbc30_test_rom` boots [ZoomTen's MBC30 test ROM](https://github.com/ZoomTen/mbc30test) (WTFPL) and screenshot-compares both verdict screens, following the `test_magen.py` pattern. On plain MBC3 the ROM reports "MBC3 SRAM OK!" instead, failing the compare.

The test ROM is fetched on demand by a new session fixture in `conftest.py`, currently from the GitHub release URL.

`pre-commit` (ruff + ruff-format) passes on all changed files.

## Verification

- ZoomTen's MBC30 test ROM reports **"MBC30 SRAM OK!"** (all 8 SRAM banks) and **"MBC30 ROM OK!"** (all 256 ROM banks); screenshots below.
- Pokémon Crystal (JP) (`PM_CRYSTAL`, cart type `0x10`, RAM size `0x05`) plays the full intro and reaches the title screen.
- No behavior change for plain MBC3.

| Crystal (JP) before | Crystal (JP) after | SRAM test | ROM test |
|---|---|---|---|
| 01-crystal-jp-before-blank | 02-crystal-jp-title-screen | 03-mbc30-test-sram-ok | 04-mbc30-test-rom-ok |


Baekalfen:

Baekalfen (approved):
Sorry for the delay. Great work! I fixed up a few things, and we're ready to merge. I haven't been able to test the Japanese ROM, but as the MBC30 tests are passing, I expect it to work.

PR #422, opened 2026-07-15
Add rtc_lock option to freeze the MBC3 RTC for deterministic emulation

fsegouin:
## Motivation

The MBC3 RTC derives its value from the host wall clock (time.time() - timezero in latch_rtc), which makes it the only source of non-determinism in an otherwise fully deterministic emulator. Any game that polls the clock can take a different execution path depending on when the emulator runs; the same ROM + save state + input sequence is not guaranteed to be bit-identical across runs.

Concrete example: Pokémon Gold/Silver/Crystal poll the RTC constantly. In an RNG-manipulation tool built on PyBoy, replaying the exact same (save state, frame delay, inputs) produced different DVs depending on the wall-clock time of the run, because a shifted second/minute boundary inside the input window perturbs the frame schedule around the game's Random calls. This breaks reproducible replay for anything Gen 2 (and more generally any RTC game) including RL training runs that assume deterministic episodes.

## Change

A new constructor kwarg, PyBoy(..., rtc_lock=True) (default False, so existing behavior is unchanged). It sets the already existing RTC.timelock flag, making latches return a frozen time instead of tracking the host clock. On cartridges without an RTC the flag is a no-op.

## Testing

`tests/test_rtc_lock.py`, black-box via the MBC register interface ($6000 latch / $4000 select / $A000 read):
- unlocked RTC honors a write to the seconds register (control case),
- locked RTC reads frozen zeros for sec/min/hour/day, even after the game writes the clock registers,
- rtc_lock=True is accepted harmlessly on a non-RTC cartridge.

### DIFF
--- a/tests/test_rtc_lock.py
+++ b/tests/test_rtc_lock.py
@@ -1,0 +1,75 @@
+#
+# License: See LICENSE.md file
+# GitHub: https://github.com/Baekalfen/PyBoy
+#
+
+import io
+from pyboy.utils import PyBoyException
+import pytest
+
+from pyboy import PyBoy
+
+
+def make_rtc_rom():
+    # Minimal MBC3+TIMER+RAM+BATTERY cartridge (0x10), 32KB RAM.
+    rom = bytearray(2 * 16 * 1024)
+    rom[0x0134:0x013C] = b"RTC TEST"
+    rom[0x0147] = 0x10
+    rom[0x0149] = 0x03
+    checksum = 0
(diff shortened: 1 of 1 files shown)


Baekalfen:

Baekalfen:
This is an experimental feature, and the same can be reproduced by setting `pyboy.rtc_lock_experimental(True)` right after the PyBoy constructor. But the tests are good, thank you!

Baekalfen (approved):

PR #436, opened 2026-09-11
Fix wave channel sample nibble order

clambro:
## Summary

Fix an audible channel 3 playback artifact caused by reading each wave RAM byte’s two samples in reverse order.

Wave RAM stores two 4-bit samples per byte, with the upper nibble played first. PyBoy currently plays the lower nibble first, reversing every adjacent pair of samples in the waveform. In Pokémon Yellow, this distortion manifests as a persistent faint, high-pitched whine over the game’s audio.

## Changes

- Correct the nibble selection in `WaveChannel.sample`.
- Add a regression test that writes `0xAB` to wave RAM, triggers channel 3, and verifies that the first exposed sample is `0xA`.

## Disclosure

The code and the above description were written by AI, but checked by me. This change resolves the audio issue as far as I can tell.


Baekalfen:
Thank you for the PR. Very interesting. I'm surprised none of the existing tests detected this, or even changed after this. I'll have to validate before merging

PR #439, opened 2026-09-18
Adding Puzzle based games wrappers.

MFaisalZaki:
# Game wrappers for Puzznic, Flipull, Amazing Tater and Adventures of Lolo

## Summary

Adds PyBoy game wrappers for four Game Boy puzzle cartridges, and folds a reverse-engineered
memory map into the existing Super Mario Land wrapper. All four games keep their board in work
RAM rather than in the background tilemap, so each wrapper reads the cartridge's own cell codes
from there and exposes them through `game_area`, with compressed and minimal mappings defined
over those codes. Each wrapper also carries what it takes to reach a chosen level, and a
`settle()` that waits for the cartridge's own definition of a finished move.

## New wrappers

- **Puzznic** (`PUZZNIC`). Reads the 12 by 10 playfield, the cursor and the block counters.
  Reaches a round either by typing its password on the title screen, read out of the
  cartridge's own table, or by hooking the stage loader. Exposes `blocks()`,
  `block_counts()`, `passwords()`, `stage_cleared()`, and a `game_over()` that reports the
  position a stage can no longer be cleared from, since blocks match in pairs and a lone block
  is stuck.
- **Flipull** (`FLIPULL`). Reads the 14 by 16 block field, the HUD counters as the digit pairs
  the cartridge stores, the CLEAR target and the clock. Selects a stage by hooking the loader
  and writing its two decimal digits, then checks the result against the stage table in ROM.
  Adds `calibrate()`, which measures off the cartridge what is not in RAM: which button throws,
  which OAM slots hold the player and the block in his hand, the pixels per row, and the hold
  that moves exactly one row.
- **Amazing Tater** (`AMAZING-TATER`). Reads the composed 20 by 18 board, the characters still
  to reach the flag, and which one the controls are on. Selects any of the 105 rooms by hooking
  the level loader and writing the index it is entered with. Exposes `taters()`, `blocks()`,
  `pits()`, `turnstiles()` with their arm masks, and `switch_tater()`, which waits out the
  lockout the cartridge applies after select.
- **Adventures of Lolo** (`LOLO2`). Combines the loaded room in work RAM with the background
  tilemap, because the board buffer is never updated after the room loads. Learns the heart,
  Framer and door tile numbers per room, since the cartridge ships eight terrain themes.
  Exposes Lolo's half-cell position, the enemies, the hearts, the Framers, the door state, and
  `rooms()`, which decodes all 163 rooms straight out of ROM bank 13.

## Super Mario Land

The existing wrapper gains Mario's state struct at `$C200` as `mario_position`,
`mario_facing`, `mario_on_ground`, `mario_speed`, `mario_direction` and `mario_jump_phase`,
the scroll registers as `camera`, and `object_slots()` over the ten-slot object array at
`$D100`. Debug-view annotations label each live object and Mario, matching what the Super
Mario Bros. Deluxe wrapper already does. No existing behaviour changes.

## Wiring, examples and tests

The four classes are registered in `manager_gen.py` and the generated manager files are
regenerated. Each game gets an example script in `extras/examples`, following the existing
scripts there. Tests write a board into work RAM by hand and check the decoding; they need no
game ROM, because they patch the cartridge title in a copy of PyBoy's default ROM so the right
wrapper is selected.

## Verification

The project compiles with Cython and the new tests pass against the compiled build. The offline
part of the suite is unchanged.

```
98 passed, 100 skipped, 1 xfailed, 1 xpassed
```

The one thing not verified is the boot route in each `start_game`, because the cartridges are
copyrighted and are not in the test set. The memory maps, the settle rules and the hold windows
come from a reverse-engineering pass over specific revisions, and each wrapper names its own in
the code.

## Files

| Area | Files |
| --- | --- |
| Wrappers | `pyboy/plugins/game_wrapper_puzznic.py`, `game_wrapper_flipull.py`, `game_wrapper_amazing_tater.py`, `game_wrapper_adventures_of_lolo.py`, and a `.pxd` for each |
| Super Mario Land | `pyboy/plugins/game_wrapper_super_mario_land.py` and `.pxd` |
| Plugin wiring | `pyboy/plugins/manager_gen.py`, `manager.py`, `manager.pxd` |
| Examples | `extras/examples/gamewrapper_puzznic.py`, `gamewrapper_flipull.py`, `gamewrapper_amazing_tater.py`, `gamewrapper_adventures_of_lolo.py` |
| Tests | `tests/test_game_wrapper_ram_boards.py` |


MFaisalZaki:
I’m sorry, but can I ask for feedback? Thanks

Baekalfen:
My feedback would be that this seems to be entirely(?) vibe coded, even the PR description. Split the PR into commits for each wrapper and write your own justification for the changes (don't make the LLM write it). Add proper tests of the individual wrappers with meaningful tests that proof the wrapper works, and will continue to work

MFaisalZaki:
Okay, but I have tested them using planners like BFWS on the ROM I have. I was working on a planning with simulators library and I thought that it would be better to make the wrappers PyBoy. That’s why I used an LLM to convert the code I had. If you think that the games are worth it. I can split them into several commits.

